PR #19, opened 2015-10-10
Update c57470761.lua

Steeldarkeagel:
Fix: Only destroys itself when the monster equipped attacks a monster and not when it is attacked.


salix5:
cecdd1d7a5f89abf5b310b271bd48501197221f9
fixed
Thank you.

PR #41, opened 2015-10-21
fix Erebus

destdev:
not sure why there's a random +1, that one screws the effect if the opponent has no cards in his hand but on the field/grave


salix5:
a2b76a897dec8388a2a7a93d07bfb6738ea72b2e
fixed
Thank you.

PR #46, opened 2015-10-22
fix grave effects return

destdev:
If a card that triggers a "when sent to the grave" effect at a different time then the actual sending (Mimicry Lanius, Dogu, Ruffian Railcar, Harpie Harpist, Evilswarm Kerykeion, Constellar Sombre, Scarm) is returned from the banished pile to the grave with cards like Burial from a Different Dimension it triggeres. That should not happen since it is returned and not sent.

Here are the fixed files for that, just adds a check to each script if it was banished bevore. I searched out every "if it was sent there this turn" card here, not sure if I was able to capture every card with this issue or if there are others. ... or if there's a better solution for this issue, if you got an idea I'd like to see it^^


nekrozar:
If returned Scarm, Malebranche of the Burning Abyss to the Graveyard by PSY-Framelord Omega's effect, Scarm didn't activate the effect. What version of YGOPRO you are using?

destdev:
hu ...
seems like I didn't test enough. Earlier today it was reported, that Burial wouldn't return cards, then it was further reported that "multiple cards" would trigger uppon being returned. After testing the 2 examples given, Mimicry Lanius (and the not yet released here Youtou - Shiranui) I asumed it'd be the case for all cards of that kind...
well, sorry then, I didn't test enough, the only card that needs to be fixed is Mimicry Lanius here, I tested all through now...
sorry for that trouble :/

nekrozar:
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=14858&keyword=&tag=-1
Q.そのターンに墓地へ送られた「RR－ミミクリー・レイニアス」が相手の発動した「D.D.クロウ」の効果によって除外されています。
このターンに「異次元からの埋葬」を発動し、その「RR－ミミクリー・レイニアス」が墓地に戻った場合、『②：このカードが墓地へ送られたターンの自分メインフェイズに、墓地のこのカードを除外して発動できる。デッキから「RR－ミミクリー・レイニアス」以外の「RR」カード１枚を手札に加える』効果を発動する事はできますか？
A.質問の状況のように、除外された「RR－ミミクリー・レイニアス」が「異次元からの埋葬」の効果で墓地へ戻ったとしても、「RR－ミミクリー・レイニアス」の効果を発動する事はできません。
（「異次元からの埋葬」の効果処理によって墓地へ戻ったカードは、『墓地へ送られた』扱いにはなりません。） 
If returned Mimicry Lanius to the Graveyard by Burial's effect, Mimicry Lanius was able to activate the effect.
I think that Mimicry Lanius is doubtful.

destdev:
uh ... I can't read that at all, but from the very broken english google translate provides the answer should mean, that you can not activate the effect and return must not be threated as sent

from google translate:
A. as status of your questions, as well as returned to the Graveyard by the effect of the excluded "RR- mimicry-Reiniasu" and "burial from another dimension", and that is to trigger the effect of "RR- mimicry-Reiniasu" **You can not**.
(Card back by the effect processing of the "burial from different dimension" to the cemetery, **must not be to treat "was sent to the Graveyard"**.)

so either google translate is even more horribly wrong than usual or you might have misstranslated something?

destdev:
thank you for your support there, with your Mimicri Lanius request everything here should be done now

PR #76, opened 2015-10-26
Chimeratech and Glad Beast Fusions

destdev:
After our last update we noticed that Chimeratech is not able to be summoned anymore. When localy testing I found this: http://imgur.com/PBkkpLF
After a little investigation and finding this commit 
https://github.com/Fluorohydride/ygopro-scripts/commit/a91364e2a90e4b21a6fc7f0f0ffdf7daed400093
I also checked the 3 Gladiator Beast Fusions edited there and they are also not be able to be summoned anymore.

Removing the parameters from the IsCanBeFusionMaterial function solves the problem. Not sure why they were there in the first place, if anyone could enlighten me please, would be very much apprechiated^^
If those are necassary I have no idea what card the function expects to be there. I've already tried inserting one or another possible card there, but then it shows another error that 2 parameters are needed, despite 2 being there.
(Side note: not sure if the other files in that commit are also causing problems, wasn't able to test that yet)

### DIFF
--- a/c27346636.lua
+++ b/c27346636.lua
@@ -38,7 +38,7 @@ function c27346636.splimit(e,se,sp,st)
 	return e:GetHandler():GetLocation()~=LOCATION_EXTRA
 end
 function c27346636.spfilter1(c,tp)
-	return c:IsCode(78868776) and c:IsAbleToDeckOrExtraAsCost() and c:IsCanBeFusionMaterial(nil,true)
+	return c:IsCode(78868776) and c:IsAbleToDeckOrExtraAsCost() and c:IsCanBeFusionMaterial()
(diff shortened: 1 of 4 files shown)


mercury233:
the core was updated in https://github.com/Fluorohydride/ygopro-core/pull/10

DailyShana:
Sorry, It is my carelessness..
It has been fixed here https://github.com/Fluorohydride/ygopro-core/commit/d1e09e5a63c79f21bccbfee5c83f5fd636fac398

destdev:
ok, I see
well, until our next update this will serve as a temporary serverside fix then until the core is updated again, thanks for the information

PR #84, opened 2015-10-27
fix

mercury233:
The opponent's pick wasn't random. Deck shuffle doesn't affect the order
of selected cards. So, show a fake pick card procedure and make it random...


mercury233:
It only happens when no shuffle is selected...

PR #86, opened 2015-10-29
fix Necrovalley

destdev:
fix this:
If you activate a monster effect that can target 1 card out of a group of itself and other cards (e.g. Trick clown, Redox, Inzector Giga-Weevil, Beast of the Pharaoh, Colossal Fighter) it is currently negated by Necrovalley even if it targets itself.

### DIFF
--- a/c47355498.lua
+++ b/c47355498.lua
@@ -80,7 +80,12 @@ function c47355498.discheck(ev,category,re,im0,im1,targets)
 	if not ex then return false end
 	if tg and tg:GetCount()>0 then
 		if targets then
-			return tg:IsExists(c47355498.disfilter1,1,nil,im0,im1,targets)
+			if targets:GetCount()==1 then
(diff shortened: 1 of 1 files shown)


salix5:
If it target a card in the graveyard, then it should be neated.

destdev:
The latest errata states:
"Negate any card effect that would **move a card in the Graveyard, other than itself**, to a different place."
In theory those card could move other cards, but in the case they are targeting themselfes they shouldn't since they try to move themselfes out of there.
Also rulings state:
- If a monster in the Graveyard can Summon itself back to the field, it DOES work.
- If a card in the Graveyard can add itself back to the hand or Deck, it DOES work.
Both rulings are from Structure Deck - Marik, should be the latest ones of that kind, or do I miss something from a later date? Also I can't seem to find anything that would qualify "targeting" to be negated overall. If you have information there, please share it, I'm not able to find one :/

PR #93, opened 2015-11-02
Update c15155568.lua

destdev:
fix: If you active the effect of Destruction Swordmaster Fusion (and propably some other cards as well) in the grave while this card is in the grave an error shows up:
[string "./script/c15155568.lua"]:42: attempt to index local 'tg' (a nil value)

the remaining change is just some eyecandy (description^^)

### DIFF
--- a/c15155568.lua
+++ b/c15155568.lua
@@ -9,17 +9,17 @@ function c15155568.initial_effect(c)
 	e1:SetTarget(c15155568.target)
 	e1:SetOperation(c15155568.activate)
 	c:RegisterEffect(e1)
-	--
-	local e4=Effect.CreateEffect(c)
-	e4:SetType(EFFECT_TYPE_FIELD+EFFECT_TYPE_QUICK_O)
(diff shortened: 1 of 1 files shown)


DailyShana:
detail situation? What card is going to be disabled?

destdev:
sorry if the explanation wasn't detailed enough^^
In the described situation this card (15155568==Destruction Sword Flash) brings up that error message, it doesn't try to activate or negate anything at all. The card itself seems to work perfectly fine, this change just gets rid of the error message.
I'm a bit puzzled myself why it tries to check for a target here, since the above mentioned effect shouldn't have one ...
...
oh god, while writing this I've just thought to check and realized that the mentioned effect has a target flag which it shouldn't have >.<
Sorry for the delayed realisation of the real problem ...

DailyShana:
I was a little lack of consideration.
In fact, checking tg is nil or not is necessary because there exists compulsory effect that needs to target cards, e.g the effect of Caius the Shadow Monarch

PR #94, opened 2015-11-03
fix Destruction Sword Flash

DailyShana:
The description of this card in wikia and yugioh-wiki is wrong, there is no restriction on the number of target cards.
The only trigger effect whick can activate in deck has been considered in Duel.IsChainDisablable


salix5:
沒注意到wiki寫錯了
謝謝

PR #105, opened 2015-11-07
fix Pandemonium

nekrozar:
Fix this: If you have Pandemonium, you must pay LP.


salix5:
It's my bad.
Thank you!

PR #128, opened 2015-11-19
fix Magical Sea Castle Aigaion

destdev:
with the "IsFacedown()" check the effect can't be activated at all, replacing it with "not IsFaceup()" works fine


mercury233:
It was fixed in
https://github.com/Fluorohydride/ygopro/commit/6c1c51cda1151eae608df584894361d3d5200b39

destdev:
kk, sorry forget it then^^

PR #166, opened 2015-12-03
fix The Phantom Fog Blade

destdev:
fix: Nothing in the card text states that the card is destroyed if the monster leaves the field (or is wikia wrong here?)


DailyShana:
そのモンスターがフィールドから離れた時にこのカードは破壊される。
When that monster leaves the field, destroy this card.

destdev:
thanks, I'll just trust you on this one and also edited wikia accordingly

PR #168, opened 2015-12-04
Update utility.lua

Tianchenglipu:
Make Auxiliary.AddFusionProcCodeFunRep & Auxiliary.AddFusionProcFunFunRep perfect for the following situations. 
·when you can use material from your oppo's side of field
·when has gc, and gc isn't on your side of field, and you have no spare MZONE
·when has gc, and gc fits code and f at the same time, or gc fits f1 and f2 at the same time


salix5:
@VanillaSalt

PR #173, opened 2015-12-06
fix Duel.XyzSummon with or without mg

Tianchenglipu:
These can now Xyz summon "Number F0: Utopic Future" or any other monsters alike it if there are any Xyz monsters can be these cards's targets.


salix5:
@VanillaSalt 
Can you help checking this part?

PR #182, opened 2015-12-17
fix ZW - Lightning Blade

DailyShana:
When this card and the equip target are going to be destroyed simultaneously, this card should be destroyed by its own effect.


DailyShana:
image

PR #189, opened 2015-12-20
fix an infinite loop

DailyShana:
When the effect of Final Attack Orders is applying, opponent special summon a monster from extra deck, then summon Night Dragolich
>先开最终突击指令，对面横着出锁链龙自动竖过来，自己通招夜龙巫妖


DailyShana:
Here is a puzzle for test.
[test puzzle-pos adjust.zip](https://github.com/Fluorohydride/ygopro-scripts/files/70407/test.puzzle-pos.adjust.zip)

salix5:
234dec4e8c129221d565aa9154047249832d643c
已修正
主要的原因應該是「設定表示形式」對象必須包含所有變更前和變更後的怪獸
這樣才能判斷最後生效的效果是哪一個

PR #193, opened 2015-12-21
fix No.38

mercury233:
Fix: The spell card negated by it will go to Overlay and then go to grave.
Fix: The atk up effect will miss timing.

### DIFF
--- a/c63767246.lua
+++ b/c63767246.lua
@@ -30,7 +30,7 @@ function c63767246.initial_effect(c)
 	e3:SetDescription(aux.Stringid(63767246,2))
 	e3:SetCategory(CATEGORY_ATKCHANGE)
 	e3:SetType(EFFECT_TYPE_FIELD+EFFECT_TYPE_TRIGGER_O)
-	e3:SetProperty(EFFECT_FLAG_CARD_TARGET)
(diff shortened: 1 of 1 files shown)


Zayelion:
#teamwork

mercury233:
@Zayelion I saw many bugs was fixed in this repository, but some varients didn't pull it. Maybe because they don't know what the modifications mean, or they don't know if this modification need core/gframe to be updated (such as Magical Sea Castle Aigaion), or they simply ignore it.

Zayelion:
We are aware of them, I am planning on Salvation pointing at a different `ygopro-script` folderon game initiation. The reason we do not usually do this is because sometimes they require `ygopro-core` updates, or `ygopro-scripts/utility.lua` updates. More so than anything updates break replays and **that highly angers users**.

We dont have a way of knowing if a fix will fubar something. Last time we pulled in new scripts from this repo it caused Lightsworn Wulf to be able to be normal summoned. We try to pull in new cards regularly, but its not like our user base is complaining about it.

If a user complains about a printed card, we always check here first per protocol.

mercury233:
@Zayelion It is so good that my users are used to face broken replays after every updates _(:з」∠)_

PR #198, opened 2015-12-24
fix Galaxy Tyranno

DailyShana:
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=11461&keyword=&tag=-1
Now No.95 can be xyz summoned via overlaying No.62 by the effect of Galaxy Tyranno


salix5:
@VanillaSalt Can you help to check these files?
Thank you!

PR #200, opened 2015-12-24
changes

Tianchenglipu:
1、EFFECT_ADD_FUSION_SETCODE
2、Can cancel Xyz summon when above 1 special material
<del>3、update for Number:21 (use aux.AddXyzProcedure2, and add param 'mc' for xyzop)</del>
this part has conflict with #198 so keep it

Another part: https://github.com/Fluorohydride/ygopro-core/pull/30


DailyShana:
How about make g:FilterSelect(tp, f, min, max, ex_card,...), etc. cancelable when min is 0?

Tianchenglipu:
Yes, that would be OK. Also, please make that change to every selective function.

Tianchenglipu:
F..k, mobile phone cannot edit comment

DailyShana:
For No.99, we cannot use cancelable selection

Tianchenglipu:
No, we can, as we can select material first then run Xyzop.

PR #201, opened 2015-12-24
Number:21 use aux.AddXyzProcedure

Tianchenglipu:
select special material first, then run xyzop
provide special material as an extra param of xyzop


Tianchenglipu:
@VanillaSalt 
Would you please check this commit? Thank you!

PR #220, opened 2016-01-11
new card SHVI

nekrozar:
http://www.db.yugioh-card.com/yugiohdb/card_search.action?ope=1&sess=1&pid=1118003&rp=99999

Blue-Eyes Spirit Dragon's script is b6af9571a0ff6a0f96b54c4baaed9a71cc5da0c8

PR #221, opened 2016-01-11
new card SHVI

mercury233:
ムーンライト: 0xdf
アモルファージ: 0xe0

Scripts are from https://github.com/Fluorohydride/ygopro-pre-script


mercury233:
@VanillaSalt Would you please check and merge this and https://github.com/Fluorohydride/ygopro-scripts/pull/220 ? I think this commit has less bugs than nekrozar's, but isn't perfect though.

PR #222, opened 2016-01-12
add English cardnames

mercury233:
Add English comments of cardnames, for non-Japanese developers.
The English names are from SalvationDevelopment.

Rewrited those scripts beacuse they use CR as line feeding.
> c39703254.lua
> c47852924.lua
> c49980185.lua
> c70946699.lua


DailyShana:
没必要吧
那我们是不是应该加入中文卡名，或者每种语言都加一遍？

StormWulf:
The English name comments makes it easier for the English-speaking developers to identify scripts without looking up the card ID. Mercury233 can add other languages in addition to Japanese and English easily if demanded.

mercury233:
实际上是@StormWulf 的建议，我实现了而已

StormWulf:
I think that it would be a great idea to include Chinese names in addition to English names to make script editing convenient for FH's team as well. It feels pointless to have Japanese name comments when most of the developers aren't Japanese speakers.

DailyShana:
They are just records, not for convenience.
If everyone wants to add comment for his convenience, it will be in a mess.

Zayelion:
I assure you it wont, we've been working with this code base for over two years (Oct 2013), we just havent been as vocal or direct. Its just now that we have become fed up with the changes in this repository negatively effecting us as various group. 

We have to pull up a website, (therefore be online), reference the id at the wikia, and then find the card to open and edit it. YGOPro itself doesnt allow copying to clipboard so using it is even more tedious. 

Being able to press control-f (find) to pull up the card in our IDE's is much more natural. It'd be a series of commit or possibly just one depending on the method used. 

Commenting your code is standard practice, as someone that programs professionally the programming being the documentation is (excuse my language) "utter bullshit". Additionally YGOPro has very few comments, or explanations on how it works, so refusing comments to clarify the intentions of the programming or even what it is, is a sin on sins.

salix5:
We do not mean to be offensive, this is just another point of view.

Basically, I believe that the conversion between ID and card name in the local language can be done by data in cards.cdb file, and this file can be edited by tools like Butty's Card Manager, 247321453/DataEditorX, or other sqlite editors offline.

Japanese is not our native language either, and I agree that the name in the script are just records. I think what @DailyShana means by "in a mess" is that the official English card names are usually not known when they are released in OCG, and if we record the temp card names in wikia or so, we'll need to change them when they are released in TCG. It seems to be an unnecessary work, since the official card names in local language is already recorded in cards.cdb file, and you can copy the ID or name from it for further use.

The lack of comments in the code is somewhat inconvenient, and we will try out best to improve this part.

PR #230, opened 2016-01-17
fix De-Synchro

destdev:
fix this: the synchro material monsters can't be summoned


salix5:
128a8fa7e755fcd1bda50ca5d074ae7253ecab0d
fixed
Thank you.

PR #255, opened 2016-02-04
fix King Tiger Wanghu

destdev:
a little something that makes his effect not working^^


salix5:
It's my mistake.
Thank you very much!

PR #258, opened 2016-02-05
fix Zaborg the Mega Monarch

destdev:
fix: if there are only monsters on the field that can't be targeted (e.g. Zaborg itself with March) you can activate the effect, this shouldn't be possible

### DIFF
--- a/c87602890.lua
+++ b/c87602890.lua
@@ -51,7 +51,7 @@ function c87602890.condition(e,tp,eg,ep,ev,re,r,rp)
 end
 function c87602890.target(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
 	if chkc then return chkc:IsLocation(LOCATION_MZONE) and chkc:IsDestructable() end
-	if chk==0 then return true end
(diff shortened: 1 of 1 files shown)


nekrozar:
Zaborg the Mega Monarch's effect is Mandatory Effects. If you control only Zaborg the Mega Monarch, Zaborg the Mega Monarch should activate.

Zaborg the Thunder Monarch's Rule:
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=4&cid=5965
■「雷帝ザボルグ」がアドバンス召喚に成功し、効果を発動する際に、フィールドに他のモンスターが１体も存在しない場合、自身を対象として効果を発動する事になります。

mercury233:
@nekrozar He means, when "進撃の帝王/March of the Monarchs" in effect and only Zaborg itself is on field. The script forgot to check if the target exists.

DailyShana:
Mandatory Effects=强制效果
It is a basic rule that a Mandatory Effect must be activated as long as you are able to pay the cost, no matter the target exists or not.

destdev:
I'm dump >.>
sorry, I forgot that it's mandatory...

PR #260, opened 2016-02-06
fix Exodius, Vennominaga, Gimmick Puppet of Leo

nekrozar:
https://github.com/Fluorohydride/ygopro/issues/1673
Fix this: Phantom of Chaos, The Tyrant Neptune and so on copy win effect.


DailyShana:
Logically, if the counters are added by other cards, the win effect of Vennominaga and Gimmick Puppet of Leo still applies.

nekrozar:
Oh, thank you.

PR #274, opened 2016-02-17
fix EFFECT_DESTROY_REPLACE

nekrozar:
Fix this: If a Blue-Eyes White Dragon you control would be destroyed by Dreadnought Dunker's effect when Blue-Eyes White Dragon would be destroyed by battle with Dreadnought Dunker, you can banish 1 Gospel of Revival instead.


DailyShana:
fixed in Fluorohydride/ygopro-core@ec0a2f8 and Fluorohydride/ygopro-core@e03c560.
Thanks.

PR #284, opened 2016-02-28
fix Destiny Hero - Dogma

destdev:
fix: the effect to halve the opponents lp should happen in the SP after the summon, no matter the way it is special summoned (thus also after being special summoned by something like "A Wild Monster Appears!" which it doesn't at the moment)


nekrozar:
If dogma Special Summoned by "A Wild Monster Appears!", Dogma doesn't activate effect.
> http://www.db.yugioh-card.com/yugiohdb/card_search.action?ope=2&cid=6757
> **この方法で特殊召喚に成功した**次の相手のスタンバイフェイズ時、相手ライフを半分にする。

destdev:
oh, "this way" ... yay for TCG screwing up the translation again >.>
thanks for the info then

PR #312, opened 2016-03-17
update Raidraptor - Nest

destdev:
the deck is automaticly shuffled after adding a card from the deck to the hand anyway
also, removing the shuffle command here prevents the deck from being shuffled when adding the card from the grave to the hand, a shuffle shouldn't happen there


salix5:
The player can confirm the deck while choosing cards, so the deck should be shuffled even if he add a card from the grave.

destdev:
oh god, my brain >.<
thanks^^

PR #314, opened 2016-03-18
errata Sangan

nekrozar:
http://www.yugioh-card.com/japan/notice/revision/#date20160401


mercury233:
it should only limit the active of card(P card), not the active of effect

salix5:
參考色彩支配者的敘述，目前先以「不能發動效果」來解讀

PR #318, opened 2016-03-19
Change timing hint hex numbers to constant

mercury233:
Make ALL SetHintTiming use constant instead of hex numbers.
Hex numbers is difficult to read and confusing to beginners.
I think `TIMINGS_CHECK_MONSTER_E` is too short, should be replaced by `TIMINGS_CHECK_MONSTER+TIMING_END_PHASE`.

Rewrited those scripts beause they are using CR+LF.
c14812659.lua
c88240999.lua
c99185129.lua


mercury233:
btw, should we let `RESET_EVENTS_COMMON`=0x1fe0000, and change `RESET_EVENT+0x1fe0000` to `RESET_EVENT+RESET_EVENTS_COMMON` ? It will make the scripts more easy to read and write.

DailyShana:
for writing, `0x1fe0000` is more convenient
for reading, "common" only indicates that it is used commonly, which is nonsense. It cannot give essential understanding.
`0x1fe0000` contains too many information to sum up as one word, so just keep it.

mercury233:
I don't think so, `0x1fe0000` is a bit difficult to remember, so it need more check if you wrote it manually.
"common" mean it is common reset event for all "add" effects, but maybe we can find a better term.

Hello23-Ygopro:
## Constants Vs Hex Values (1/2)

Its a shame this was closed.. 

@mercury233 i agree , with using Constants instead of Hex values in situations like this , where the Hex Values are either too long , or difficult to remember
Hex Values in general need more information and more memorization
Constants make more sense , regardless of scripters skill/experience , and is more convenient to use in scripting

### Theres alot of benefits using more Constants over Hex Values in general : 
1) In general , Words (Constants) make more sense then Numbers and Letters in combination (Hex),
2) if the Hex Values change in the future , they just have to edit the Hex values inside constant.lua file , and they dont have to modify all 3000+ scripts individually 
3) it makes scripting easier and more convenient , see 1.
4) it lessens workload and saves time , see 2.
5) this can attract more beginners to help with scripting in case of short handedness , since it can appear as being less complicated if more Constants are used
(personally , when i started scripting , the Reset Hex Values were harder to learn them most other features)
##
I hope the people in charge here reconsider using more Constants in general in the near future
not everything has to use Constants , just at least the Resets like @mercury233 suggested , plus anything that has long complicated Hex Values

Hello23-Ygopro:
## Constants Vs Hex Values (2/2)

I also have a few suggestions
i think i made some of them in the past , and i saw some people have the same ideas
#
### 1) Adding more Constants for popular Card Names , for example :
```
--========== Card Names ==========
CARD_POLYMERIZATION=24094653
CARD_BLUE_EYES_SPIRIT=59822133
...ETC ..
```

### 2) Adding Constants for SetCodes would help alot , for example : 
```
--========== SetCodes ==========
SET_ALLY_OF_JUSTICE=0x1
SET_GEX=0x2
...ETC ..
```
### 3) Adding Constants for Counters , for example :  
```
--========== Counters ==========
COUNTER_SPELL=0x1
COUNTER_WEDGE=0x1002
...ETC ..
```

### 4) Some more Constants that have less priority : 
There is already a `RACE_ALL		=0xffffff		--全种族`
Maybe a `LOCATION_ALL=0ff`
In situations like "Macro Cosmos"..
and `ATTRIBUTE_ALL=0xff`
...ETC ..

The only downsides of more Constants I can think of : 
- -The "constant.lua" file would constantly have to update , more updated then usual ,
+but you can still use the archetypes Hex values if the scripter prefers ,
- -This can increase the file size of "constant.lua" if alot of Constants are added more frequently ,
+but it shouldnt be so big that it would be a problem

### 5) An idea replacing `c12345678.`(Script name) & `12345678`(Code) inside each script :
```
local function getID()
	local str=string.match(debug.getinfo(2,'S')['source'],"c%d+%.lua")
	str=string.sub(str,1,string.len(str)-4)
	local scard=_G[str]
	local sid=tonumber(string.sub(str,2))
	return scard,sid
end
local scard,sid=getID()
```
scard = the source "c number"/script name, for example "c12345678."
sid = the source"id number"/code for example "12345678"

This script has been used in the past on Custom Cards , in case the scripts name/cards ID number changes , you dont have to modify anything in each script so its also less work

I dont think this is useful/necessary on official archived scripts , because official card numbers usually dont change at all , but in [Pre-Scripts](https://github.com/Fluorohydride/ygopro-pre-script) and Custom Cards its most useful , since their card numbers change alot

Thank You!

PR #321, opened 2016-03-21
fix Performage Hat Tricker

destdev:
fix this: if you activate the effect several times in the same chain and fail to add a counter due to the max amount being reached somewhen in the middle of the resolution the damage is still negated (example: you have 4 effects that would inflict damage, you activate the effect 4 times -> the 4th effect would still deal 0 damage, despite no counter being placed)


salix5:
f2b9f9544ee6c391c9c3c708b4736803aac71a1a
fixed
Thank you.
The return value of AddCounter() is a bolean value.

destdev:
so it's actually that simple Oo
thanks for the explanation

PR #322, opened 2016-03-22
fix The Dominion of the Legendary Monarch

destdev:
fix this: If P1 controls Dominion with an empty extra deck and P2 has only 1 tribute summoned monster on the field that tribute summoned monster can be used as a material for an extra deck summon. However since the monster is no longer on the field in the process the summon stops working in the middle of the procedure.
Thus you shouldn't be able to select that tribute summoned monster in the first place.

### DIFF
--- a/c84171830.lua
+++ b/c84171830.lua
@@ -34,6 +34,22 @@ function c84171830.initial_effect(c)
 	e4:SetTarget(c84171830.lvtg)
 	e4:SetOperation(c84171830.lvop)
 	c:RegisterEffect(e4)
+	--prevent be material
+	local e5=Effect.CreateEffect(c)
+	e5:SetType(EFFECT_TYPE_FIELD)
+	e5:SetCode(EFFECT_CANNOT_BE_FUSION_MATERIAL)
+	e5:SetRange(LOCATION_FZONE)
+	e5:SetTargetRange(0,LOCATION_MZONE)
+	e5:SetCondition(c84171830.matcon)
+	e5:SetTarget(c84171830.mattg)
+	e5:SetValue(1)
+	c:RegisterEffect(e5)
+	local e6=e5:Clone()
+	e6:SetCode(EFFECT_CANNOT_BE_SYNCHRO_MATERIAL)
+	c:RegisterEffect(e6)
+	local e7=e6:Clone()
+	e7:SetCode(EFFECT_CANNOT_BE_XYZ_MATERIAL)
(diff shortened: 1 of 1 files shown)


salix5:
真帝王領域
①：自分のエクストラデッキにカードが存在せず、自分フィールドにのみアドバンス召喚したモンスターが存在する場合、相手はエクストラデッキからモンスターを特殊召喚できない。
②：自分のアドバンス召喚したモンスターの攻撃力は、相手モンスターに攻撃するダメージ計算時のみ８００アップする。
③：１ターンに１度、自分メインフェイズにこの効果を発動できる。自分の手札の攻撃力２８００／守備力１０００のモンスター１体を選び、そのモンスターのレベルをターン終了時まで２つ下げる。

Domain of the True Monarchs
While you have no cards in your Extra Deck and you are the only player that controls a Tribute Summoned monster, your opponent cannot Special Summon monsters from the Extra Deck. If a Tribute Summoned monster you control attacks an opponent's monster, it gains 800 ATK during damage calculation only. Once per turn: You can reduce the Level of 1 monster with 2800 ATK and 1000 DEF in your hand by 2 until the end of this turn (even after they are Summoned/Set, and even if this card leaves the field).

P1: Domain, extra=0
P2: tribute summoned monster A

In this case Domain has no effect since P1 does not control a tribute summoned monster.

destdev:
So you can use the monster as material even if it's certain to fail?
I thought that stuff that is certain to fail needs to be prevented in the first place.
If so, ok, but then there is another problem:
If an xyz summon fails the material kinda stays on the field in a limbo state.
dominion
A xyz Material without an xyz monster floating around, it can't do anything and the opponent can attack through it.

destdev:
I investigated and tested a bit further now.
Unfortunately I couldn't find anything in rulings regarding this issue.
Taking a similar situation as in the picture however I found out, that you are currently perfectly able to xyz summon with 2 (tribute summoned) Red Layers.
dominion
so, either way, something is wrong here, but I'm currently unsure what exactly (and I've noticed a huge mistake on my part on this commit, have to correct that later :/)
... either what I think, Dominion itself that currently doesn't prevent the last tribute summoned Monster to be used as a material
... or it is as you say, but then the xyz procedure in utility has to be edited in a way that shouldn't allow the summon in this case and sends the material to the grave
... there also might be an error with the Super Quantum field spell, since the summon with that procedure leaves the material as shown in my previous comment

PR #328, opened 2016-03-30
fix Interrupted Kaiju Slumber

destdev:
fix, sg must be defined after the destruction
else you can't special summon if something like Rivalry of Warlords is chained to Slumber since the value before the destruction is 0 but becomes valid after that


DailyShana:
fixed in 009fc6118a15a59ca085fb2fc1f6be11fe6686bf in principle

PR #342, opened 2016-04-05
fix Cyber Valley

destdev:
if not both cards are banished you shouldn't be able to draw (e2) or able to return the card to the deck (e3)
you are currently able to draw, even if banishing fails completely (e.g. Iron Wall is chained to the effect)


destdev:
oh sorry, my bad ...
tested only if the effect still works and missed the part where the fix should work, I thought it should >.<
yea, putting () around works, thanks for the explanation

PR #357, opened 2016-04-11
fix Master Peace

destdev:
The effect should only negate the activation like LaDD, not destroy the card


mercury233:
http://www.db.yugioh-card.com/yugiohdb/card_search.action?ope=2&cid=12419
It said 
> その発動を無効にし破壊する。

destdev:
oh please, don't tell me wikia fucked up even more than it usually does >.<
ok then, thanks

PR #369, opened 2016-04-17
fix Orea, the Sylvan High Arbiter

destdev:
the card text says
"[...]equal to the Level that the sent monster had in the hand or on the field[...]"
but at the moment the level is taken from the value in grave


DailyShana:
> Ｑ：カードをめくる枚数は墓地に送った植物族モンスターのどの場所でレベルを参照しますか？
> Ａ：墓地でのレベルを参照します。(14/02/15)

It says that the level is taken from the value in grave
http://yugioh-wiki.net/?%A1%D4%BF%B9%CD%E5%A4%CE%C4%C3%BF%C0%20%A5%AA%A5%EC%A5%A4%A5%A2%A1%D5

destdev:
ok, thank you, another case of TCG fucking up the translation >.<

PR #381, opened 2016-04-22
add hint for 爆竜剣士イグニスターP

mercury233:


DailyShana:
This two effects can be distinguished by observing whether some card on field is targeted.

mercury233:
参考圣灵兽骑 雷鹰，我感觉还是加上比较合适

mercury233:
I know this change isn't neccessary, but I think it will make the game better. What do you think?
@salix5 @VanillaSalt @woodee

PR #384, opened 2016-04-24
fix Master Peace

destdev:
fix this: if Master Peace is equipped to a monster and that monster leaves the field you are currently able to activate Master Peace effect to special summon


destdev:
fixed with https://github.com/Fluorohydride/ygopro-scripts/commit/4b1cad5517e638551d40b84331e33bdaacce909d

PR #387, opened 2016-04-25
fix Vampire Sorcerer

destdev:
fix this: Vampire Sorcerer is currently able to activate its effect if tributed for a tribute summon with an opponents The Monarchs Stormforth
a tribute due to Stormforth is not considered a removal due to a card effect


DailyShana:
It only says "If this card is sent to Graveyard by your opponent"(このカードが相手によって墓地へ送られた場合) in Japanese.

destdev:
if I wouldn't be so horrible in learning foreign languages I'd consider learning japanese someday instead of relying on translations >.<
anyway, thanks for the clearification

PR #403, opened 2016-05-06
Update utility.lua

Tianchenglipu:
To avoid material_count, fit_monster and other special features being copied by Card.CopyEffect

### DIFF
--- a/utility.lua
+++ b/utility.lua
@@ -427,6 +427,7 @@ end
 --material_count: number of different names in material list
 --material: names in material list
 function Auxiliary.AddFusionProcCode2(c,code1,code2,sub,insf)
+	if c:IsStatus(STATUS_COPYING_EFFECT) then return end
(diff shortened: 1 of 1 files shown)


Tianchenglipu:
@VanillaSalt @salix5 
I have "Cybernetic Hidden Technology"(92773018) in my LOCATION_SZONE face-up and three "Phantom of Chaos"(30312361) in my LOCATION_MZONE face-up, banish a "Cyber End Dragon"(1546123) in my LOCATION_GRAVE to activate the effect of  "Phantom of Chaos". After this, I end my turn. 
At the next turn, when my opponent tries to attack me, I can cost "Phantom of Chaos" to activate the effect of "Cybernetic Hidden Technology" to destroy the attacking monster. Why? Because "c1546123.material" has been copied by "Phantom of Chaos". This is incorrect.

PR #420, opened 2016-05-22
update Horakhty

destdev:
For whatever reason people want to try and summon Horakhty with the new card Lullaby of Obedience now ...

According to the card text Horakhty should be able to grant the summon player (thus, the opponent of the player who activates Lullaby) the win in this case if I'm not mistaken. The win condition must be moved outside the summon procedure in this case.


DailyShana:
You should use `EFFECT_TYPE_SINGLE`

destdev:
done
any reason for that not to use EFFECT_TYPE_FIELD like Ghostrick Angel of Mischief?

DailyShana:
There is no clear timing for the winning effect of _Ghostrick Angel of Mischief_.

PR #434, opened 2016-05-31
fix Brotherhood of the Fire Fist - Buffalo

mercury233:
Fix: It should be able to spsummon when the monster zone is full.


destdev:
this should only work for a location count of -1 or greater, if it is -2 or lower (possible with something like Kaiser Colloseum) it shouldn't be able to be special summoned
for chk, ct should return <3 if I'm not mistaken

mercury233:
@destdev Thanks. Fixed.

PR #461, opened 2016-06-25
fix Storm Neos + Chaos Neos

destdev:
similar to the fix https://github.com/Fluorohydride/ygopro-scripts/commit/dca7df498bbd8d6ce7e62a759cd307981ca25722
at the moment you can summon Storm Neos if you have neos + Air Hummingbird on the field only and use Fusion Tag on Air Hummingbird and reveal Marine Dolphin (also no cards are returned to the deck then, you can simply summon 3 Storm Neos)
same situation with Chaos Neos and Fusion Tag on Dark Panther by revealing Twinkle Moss

### DIFF
--- a/c17032740.lua
+++ b/c17032740.lua
@@ -58,19 +58,19 @@ function c17032740.spfilter1(c,mg,ft)
 	local mg2=mg:Clone()
 	mg2:RemoveCard(c)
 	if c:IsLocation(LOCATION_MZONE) then ft=ft+1 end
-	return ft>=-1 and c:IsFusionCode(89943723) and c:IsAbleToDeckOrExtraAsCost() and c:IsCanBeFusionMaterial()
+	return ft>=-1 and c:IsFusionCode(89943723) and c:IsAbleToDeckOrExtraAsCost() and c:IsCanBeFusionMaterial(nil,true)
 		and mg2:IsExists(c17032740.spfilter2,1,nil,mg2,ft)
 end
 function c17032740.spfilter2(c,mg,ft)
 	local mg2=mg:Clone()
 	mg2:RemoveCard(c)
 	if c:IsLocation(LOCATION_MZONE) then ft=ft+1 end
-	return ft>=0 and c:IsFusionCode(43237273) and c:IsAbleToDeckOrExtraAsCost() and c:IsCanBeFusionMaterial()
+	return ft>=0 and c:IsFusionCode(43237273) and c:IsAbleToDeckOrExtraAsCost() and c:IsCanBeFusionMaterial(nil,true)
 		and mg2:IsExists(c17032740.spfilter3,1,nil,ft)
 end
 function c17032740.spfilter3(c,ft)
 	if c:IsLocation(LOCATION_MZONE) then ft=ft+1 end
-	return ft>=1 and c:IsFusionCode(17732278) and c:IsAbleToDeckOrExtraAsCost() and c:IsCanBeFusionMaterial()
+	return ft>=1 and c:IsFusionCode(17732278) and c:IsAbleToDeckOrExtraAsCost() and c:IsCanBeFusionMaterial(nil,true)
 end
 function c17032740.spcon(e,c)
 	if c==nil then return true end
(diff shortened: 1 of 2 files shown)


c17032740.lua
```diff
@@ -51,49 +51,80 @@ end
...
-	local f3=g3:FilterCount(Card.IsLocation,nil,LOCATION_MZONE)>0 and 1 or 0
-	if ft==-2 then return f1+f2+f3==3
-	elseif ft==-1 then return f1+f2+f3>=2
-	else return f1+f2+f3>=1 end
+	local mg=Duel.GetMatchingGroup(c17032740.spfilter,tp,LOCATION_ONFIELD,0,nil,89943723,43237273,17732278)
+	if ft>0 then return mg:IsExists(c17032740.spfilter1,1,nil,mg) end
+	local f1=Duel.GetMatchingGroupCount(c17032740.spfilter,tp,LOCATION_MZONE,0,nil,89943723,0,0)>0 and 1 or 0
+	local f2=Duel.GetMatchingGroupCount(c17032740.spfilter,tp,LOCATION_MZONE,0,nil,43237273,0,0)>0 and 1 or 0
+	local f3=Duel.GetMatchingGroupCount(c17032740.spfilter,tp,LOCATION_MZONE,0,nil,17732278,0,0)>0 and 1 or 0
+	if ft==-2 then return f1+f2+f3==3 and mg:IsExists(c17032740.spfilter1,1,nil,mg)
```
DailyShana: f1,f2,f3 maybe indicate the same monster.

destdev:
this should do the job

c17032740.lua
```diff
@@ -51,49 +51,79 @@ end
...
-	local f1=g1:FilterCount(Card.IsLocation,nil,LOCATION_MZONE)>0 and 1 or 0
-	local f2=g2:FilterCount(Card.IsLocation,nil,LOCATION_MZONE)>0 and 1 or 0
-	local f3=g3:FilterCount(Card.IsLocation,nil,LOCATION_MZONE)>0 and 1 or 0
-	if ft==-2 then return f1+f2+f3==3
-	elseif ft==-1 then return f1+f2+f3>=2
-	else return f1+f2+f3>=1 end
+	local mg=Duel.GetMatchingGroup(c17032740.spfilter,tp,LOCATION_ONFIELD,0,nil,89943723,43237273,17732278)
+	if ft>0 then return mg:IsExists(c17032740.spfilter1,1,nil,mg) end
+	local mg2=mg:Filter(Card.IsLocation,nil,LOCATION_MZONE)
+	local ct=mg2:GetClassCount(Card.GetFusionCode)
```
DailyShana: Since `Card.GetFusionCode` has multiple return value, it's not suitable to use `GetClassCount`, which only compare the first return value.

destdev:
I hope I didn't miss anything else^^
I also noticed a bug while testing, IsCanBeFusionMaterial() always fails with cards in the SZONE, you might want to change that.
so, what did I do here? - I couldn't think of a better way to check that, now the ft check is within the spfilter functions. Thus that part of the condition is gone now, also the check for that in the operation is no longer necassary.
I also noticed another bug with the removal of fusion codes from the group, since cards can have multiple of those I always came across errors when using Fusion Tag in one way or another. Thus now only tc is removed.
I also fixed a little bug that increased ft, even if you selected a card from the s/t zone.

DailyShana:
`IsCanBeFusionMaterial(nil,true)` will allow trap/spell used as fusion material.
_Luster Pendulum, the Dracoslayer_ can get the name of _Neos_ by _Hero Mask_ for instance.

destdev:
thanks, no idea why I didn't get that myself ... just looking at the function makes it obvious, fixed now^^

I know that there are several ways to change the code and thus the fusion code of any card these days. However removing the whole Fusioncode from the card pool you choose from can cause several errors, the script fails again and we are back to the square one like it is now: Error Message and you can just summon 3 copies with nothing returned...
I tried to eliminate that Error with a function that not only checks for the Fusioncode but also the actual code, that one at least prevented a Fusion Taged Air Humingbird from being removed after selecting the original Dolphin. However I also stumbled uppon this example 2 days ago: Have 2 Neos and 1 Air Humingbird, use Fusion Tag on 1 Neos to add the Aqua Dolphin Fusion Code. By attempting to summon Storm Neos now you select the other Neos. By removing only the selected card the function works properly. However by removing the whole Fusioncode you also remove the other Neos from the card pool and you get said error. Thus I reverted that thing and changed it to only remove the selected card.
Unless I'm missing something, just removing the selected card and not the whole code pool doesn't cause any errors, each code is only requested once anyway. Else I'd like to get suggestions how to remove the whole Fusioncode without causing errors, I'm out of ideas now...

PR #481, opened 2016-07-07
fix Dark Sanctuary

nekrozar:
If player have no space in monster or spell&trap zone, player don't place Spirit Message.


mercury233:
It should be forced to spsummon if the player have no space in S/T zone.

PR #482, opened 2016-07-08
fix Appointer of the Red Lotus

destdev:
fix: If you activate the card during your opponents turn outside of the end phase the card currently returns on the same turn. It should only return during the end phase of the opponents next turn.


DailyShana:
>次の相手のエンドフェイズ時まで、選択したカードをゲームから除外する。

The Japanese text only says that until the next opponent's End Phase, remove the selected cards from game.

destdev:
I hate to be unable to read those things without a (usually bad) translator >.<
thanks for clearification

DailyShana:
>Remove that card from play until the End Phase of your opponent's next turn.

I think maybe the effect of this card is different in TCG.

PR #498, opened 2016-07-21
fix 真竜皇バハルストスF

mercury233:
Fix: It should remove spell & traps.


DailyShana:
Oh, it should be location on field

PR #500, opened 2016-07-21
fix Black Garden

destdev:
fix: if Remove Brainwashing is on the field, tokens summoned to the field of the controller of black garden currently switch control to the opponent. However, the owner of the tokens should be tp in any case since he summons those, not the opponent.


DailyShana:
>http://yugioh-wiki.net/index.php?%A1%D4%A5%D6%A5%E9%A5%C3%A5%AF%A1%A6%A5%AC%A1%BC%A5%C7%A5%F3%A1%D5=
Ｑ：ローズ・トークンの元々の持ち主はどちらのプレイヤーですか？
Ａ：ローズ・トークンの元々の持ち主はローズ・トークンの特殊召喚のトリガーになったモンスターを召喚・特殊召喚したプレイヤーになります。(11/09/25)

The owner of the _Rose Token_ is the controller of the monster whose summon/spsummon triggered the the creation of the token.

destdev:
hu ... ok then, didn't know that one, thanks
I always thought the owner would be the one summoning them, which is the owner :/

PR #508, opened 2016-07-26
fix Transmodify

nekrozar:
Fix this: If the monster with the same Type as sent Crystal Beast in the Graveyard is not exist in Deck except DARK monster while Advanced Dark has on the field, Transmodify cannot activate.


destdev:
you should also check for the CB Setcode, none CB don't become dark

nekrozar:
Sorry, I mistook.
https://github.com/Fluorohydride/ygopro-scripts/pull/508/commits/067075329ea576cf0cb74ee0bac2b4d0576d72b5 Fixed.

PR #516, opened 2016-08-04
fix Ancient Telescope

nekrozar:
Fix 1: Player cannot confirm 4 or less cards from the top of opponent Deck.
Fix 2: Player cannot activate Ancient Telescope if opponent have 4 or less cards in opponent's Deck.

http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=4&cid=4852
■確認するカードの枚数は効果処理時に1～5枚を選びます。
■相手のデッキの枚数が5枚未満の場合でも、相手のデッキが0枚の状態でなければ、「古代の遠眼鏡」を発動する事はできます。

### DIFF
--- a/c17092736.lua
+++ b/c17092736.lua
@@ -10,12 +10,18 @@ function c17092736.initial_effect(c)
 	c:RegisterEffect(e1)
 end
 function c17092736.cftg(e,tp,eg,ep,ev,re,r,rp,chk)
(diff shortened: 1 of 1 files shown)


DailyShana:
41aee367319b8d4f5c7cd429ef552140
Maybe there is sth wrong with database because the text is not "at most 5 cards".

PR #554, opened 2016-08-21
New auxiliary functions from Ygopro Percy

EerieCode:
These two functions are currently used in the latest version of Ygopro Percy: I figured it would be nice to have them here as well. A brief explanation:
- Card.IsSummonType: this is a shortcut for effects that activate when a card is or has been Fusion Summoned, Synchro Summoned, etc. (ex. "Starve Venom Fusion Dragon"). It's a little easier on the eyes than the full version currently in use, it might make future scripts a little faster to write.
- Auxiliary.bfgcost: Short for "Banish from Graveyard", it condenses in one single line the cost "You can banish this card from your Graveyard; ": since pretty much all archetypes nowadays have a card with this effect, I figured it'd be nice to have it already prepared in the utilities, like with Auxiliary.exccon, instead of having to write it manually in every single script.

### DIFF
--- a/utility.lua
+++ b/utility.lua
@@ -1722,3 +1722,12 @@ end
 function Auxiliary.nvfilter(c)
 	return not c:IsHasEffect(EFFECT_NECRO_VALLEY)
 end
+--Function to check the summon method used for the card. Credit goes to Cute-Nekomimi
+function Card.IsSummonType(c,t)
+	return bit.band(c:GetSummonType(),t)==t
+end
+--Cost for effect "You can banish this card from your Graveyard"
+function Auxiliary.bfgcost(e,tp,eg,ep,ev,re,r,rp,chk)
+	if chk==0 then return e:GetHandler():IsAbleToRemoveAsCost() end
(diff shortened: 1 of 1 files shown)


mercury233:
I've added `Card.IsSummonType` to ocgcore : https://github.com/Fluorohydride/ygopro-core/commit/0a53f9690c737e64ca578198be425478843b3ad7
Can you update this pull request and make all cards (about 196) use `aux.bfgcost` ?

EerieCode:
It shouldn't be a problem, though it might take a while. I should have them all ready by the weekend.

mercury233:
should we also update kozmo monsters and rescue rabbit etc.?

EerieCode:
Github was giving me problems when trying to update this branch, so I made [a new one](https://github.com/Fluorohydride/ygopro-scripts/pull/878). This one can be safely closed, all new functions from this utility are already there.
As for updating cards like Rescue Rabbit, this function should be usable for those too: the branch I made has only the ones that banish themselves from the Graveyard, though, so I'll have to upload them another time.

PR #562, opened 2016-08-28
fix Blue-Eyes Spirit Dragon related cards

mercury233:
Fix: Some cards shouldn't ignore Blue-Eyes Spirit Dragon, and some cards should be able to spsummon 1 monster instead of 0 when Blue-Eyes Spirit Dragon is in effect.


salix5:
關於精靈龍在場上時「盡可能特殊召喚」應該如何處理這點有參考資料嗎？

mercury233:
qq 20160829002222

PR #563, opened 2016-08-28
fix Exodius

destdev:
the removal of the delay flag from the last commit https://github.com/Fluorohydride/ygopro-scripts/commit/33a1a01618fa80f133f88ff22c3527cca7916b43 causes the win effect to not work anymore

no idea why (and no idea if the other cards edited there are also affected, I don't have time to test that atm)


salix5:
Do you have a replay file or video?
Thank you in advance!

destdev:
hm, I just noticed something
it doesn't grant the win uppon sending the last piece of exodia in the grave, but it grants the win after I attack another time and send anything else
here is a replay, the last piece is sent turn 5, after sending something turn 7 the win is granted

[2016-08-28 18-40-00.zip](https://github.com/Fluorohydride/ygopro-scripts/files/441233/2016-08-28.18-40-00.zip)

DailyShana:
because after `Duel.SendtoGrave(tc,REASON_EFFECT)` winop is excuted immediately before `tc:CreateRelation(c,RESET_EVENT+0x1fe0000)`

DailyShana:
special winning should not be checked during chain solving so `EFFECT_FLAG_DELAY` is necessary.

PR #571, opened 2016-09-08
fix Card of Demise

destdev:
it shouldn't count itself among the cards in the hand, this causes a few problems:
- you can't activate it if you have 3 cards including itself in the hand
- you can activate it if you have only 1 card in your deck and 2 cards including itself in the hand)

### DIFF
--- a/c59750328.lua
+++ b/c59750328.lua
@@ -22,8 +22,9 @@ function c59750328.cost(e,tp,eg,ep,ev,re,r,rp,chk)
 	Duel.RegisterEffect(e1,tp)
 end
 function c59750328.target(e,tp,eg,ep,ev,re,r,rp,chk)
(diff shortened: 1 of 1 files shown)


DailyShana:
0b8bc1c881f6709751e2d489e005a9ea
Konami's answer: you can't activate _Card of Demise_ if you have 3 cards including itself in the hand

destdev:
oh ok, thanks
however I still think that you shouldn't be able to activate it if you can't draw, just like any other draw effect
you are able to do this at the moment:
demise
I made another edit now that solves this. Or is there another ruling that would allow this?

DailyShana:
image
The answer is changed: you can activate Card of Demise if you have 3 cards including itself in the hand.
Another question: you cannot activate Card of Demise that is the only card in hand and there are only 2 cards in deck.

DailyShana:
b98eea5f4f290e1bd6043efb5d5483e7bf4e4241

PR #598, opened 2016-10-06
update Scrap Recycler (fixed)

Hello23-Ygopro:
added missing category


DailyShana:
send to deck is cost, category only records effect.

PR #601, opened 2016-10-07
fix Azure-Eyes Silver Dragon

Hello23-Ygopro:
different string texts for 2 different effects
important: might need to change database in ygopro ads and other ygopros too !


DailyShana:
ee84f2441724d935357fb2ff451cc2ed6c5c35d7
Thank you

PR #604, opened 2016-10-09
fix Interrupted Kaiju Slumber

nekrozar:
Fix this: If player cannot be used all Monster Zone by _Zany Zebra_ and _Ground Collapse_, player can activate _Interrupted Kaiju Slumber_.


DailyShana:
722c213be5e5bbbbd4f61c4652a5a64a0c63ac64

PR #616, opened 2016-10-20
fix Raremetalfoes Bismugear

Hello23-Ygopro:
fixed stringid

note : cards.zh-CN.cdb (Ygopro Chinese MyCard, i think) , has no strings texts for Raremetalfoes Bismugear[18716735] in their database
yugioh continues to evolve , more complecated effects are introduced , string texts will become more important , and shouldn't be ignored , especially with more " Copy effects " existing now


DailyShana:
ee84f2441724d935357fb2ff451cc2ed6c5c35d7
Thank you

PR #619, opened 2016-10-23
fix Rank-Up-Magic Admiration of the Thousands

Hello23-Ygopro:
fixed stringid

There was no 96142517,0 before , it was skipped
Note : cards.zh-CN.cdb & cards.ja-JP.cdb have string texts for both 96142517,0 & 96142517,1 , so this edit doesnt conflict with those databases


mercury233:
I think you should set the desc for the activate instead. The official db in Chinese do have str0 and str1 for this card.

Hello23-Ygopro:
@mercury233 , thanks , i added str0 to the first effect

the reason i didn't put str0 for the special summon effect is cuz i usually don't see any strings on EFFECT_TYPE_ACTIVATE effect types , even those with multiple effects sometimes don't have any strings

note : because str0 is now present in the first activate effect , some databases might need to add "Special Summon"
i saw Japanese Ygopro have str0 = "対象の指定を続けますか？" & str1 = "対象の指定を続けますか？"

PR #621, opened 2016-10-24
fix Gladiator Beast Nerokius

Hello23-Ygopro:
fixed stringid

29357956,0 was also skipped here

Some English databases show str0 translated as "can't activate effects", which is its Continous Effect , but in my experience Continous Effects generally don't use string texts anymore , only in cases of `EFFECT_FLAG_CLIENT_HINT` , where it needs to display the effect to the user , or when multiple Continous Effects are applied at the same time in the same effect , which is very rare


Hello23-Ygopro:
Closed , not important

PR #622, opened 2016-10-24
fix Kunoichi

destdev:
fix: in some circumstances this card discards from the opponent of the play activating the effect, this shouldn't happen

Example of such a situation:
[2016-10-24 14-49-32.zip](https://github.com/Fluorohydride/ygopro-scripts/files/549025/2016-10-24.14-49-32.zip)

I can't see any rulings posted in the Q&A of ygodb at the moment, but at least rulings found on wikia (http://yugioh.wikia.com/wiki/Card_Rulings%3AKunoichi) suggest, that the player who controls the card upon activation should always be the player discarding


DailyShana:
https://github.com/Fluorohydride/ygopro-core/commit/7f7633b1e5309a97918c8f7e302da007ae1b71cb

PR #634, opened 2016-10-28
update Goyo Defender

Hello23-Ygopro:
I noticed the Ignition Effect doesn't have a string
In my experience , most of effects that start a Chain Link have strings and also those that happen at the same time
Its not that important , but just to be safe when another effect that will copy the Ignition Effect , it will show up as "???"


Hello23-Ygopro:
Closed , not important

PR #675, opened 2016-11-12
add desc for Kozmo Sliprider

mercury233:
Chain 1 Dark Hole, chain 2 spsumon this card, both effect will be able to activate, so we need some desc.


YGO5upp0r73r:
@mercury233 Arigatou.

Although, I would recommend adding descriptions to anything that starts a Chain Link from now on when scripting new cards, just to be on the safe side.

PR #678, opened 2016-11-14
fixed corrupted keys

edo9300:
the token were destroyed even if the monster was face down


DailyShana:
>選択したモンスターがフィールド上から離れた時、この効果で特殊召喚したトークンを全て破壊する。

the Japanese text only says that "When the selected monster leaves the field, destroy all the tokens that are special summoned by this effect."

I don't know if there is some FAQ requires the monster is face up when leaving field.

edo9300:
@DailyShana http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=12774&request_locale=ja

DailyShana:
It only says that if the target monster become face down, the tokens will not be destroyed.

PR #684, opened 2016-11-20
fix 召喚獣メルカバー

mercury233:
Fix: It should only check TYPE_MONSTER+TYPE_SPELL+TYPE_TRAP, otherwise it will be able to discard continuous spell to negate continuous trap.


nekrozar:
Same problem exist in _Ultimate Providence_.

PR #705, opened 2016-12-11
fixed monsters with the possibility to be tribute summoned with only 1 tribute

edo9300:
These monsters become bugged allowing you to "tribute summon" them in
this situation:
Your monster zones are full, with not valid tributes (e.g. Summoner
Monk)
Your opponent has a valid tribute on their field
You activate The Monarch Stormforth or Soul Exchange
Now you're able to make a pseudo-tribute summon, but the monster will
remain in the hand and it can be summoned again and, the most important
thing, you can activate its effects when summoned. I didn't fix fog king
because it uses a different method, but it's still bugged.


mercury233:
i didn't test, but did you noticed this? https://github.com/Fluorohydride/ygopro/issues/1852

edo9300:
@mercury233 i didn't XD. Anyway that commit doesn't fix every monster, as you can see there only 24 cards were fixed and, now that i see them, i think that the filter in the operation is useless, because the functin already filters. Pnyway this is a core problem too, not only a script one, as i said for fog king

PR #712, opened 2016-12-19
Added Auxiliary.AddFusionProcFunMulti()

EerieCode:
The function I added to the utilities is a procedure that allows the Fusion Summon of a monster that requires any fixed number of non-named monsters as Fusion Materials (the card that led to me writing this was "[Supreme King Dragon Zarc](http://yugioh.wikia.com/wiki/Supreme_King_Dragon_Zarc)", who requires 4 non-named Materials. The function is generic, and can work regardless of the number of Materials. The script for Zarc using this procedure can be found [here](https://raw.githubusercontent.com/EerieCode/EC3/master/Zarc.lua), it's synchronized with the version currently on ygopro-pre-script. I've tested that script in various puzzle duels, using different methods such as "Polymerization", "Super Polymerization", "Miracle Synchro Fusion" and "Chain Material", and using various combinations of Pendulum hybrids such as "Odd-Eyes Raging Dragon" and "Clear Wing Fast Dragon": the Fusion Summon happened without issues in all those circumstances. Feel free to test it yourself, but I think adding this might prove to be useful in case more complex Fusion Monsters come out in the future.

### DIFF
--- a/utility.lua
+++ b/utility.lua
@@ -1333,7 +1333,12 @@ function Auxiliary.FConditionFilterMulti(c,mg,funs,n,tbt)
 	return false
 end
 function Auxiliary.CloneTable(g)
-	return {table.unpack(g)}
+	local ng={}
+	for i=1,#g do
+		local sg=g[i]:Clone()
+		table.insert(ng,sg)
+	end
+	return ng
 end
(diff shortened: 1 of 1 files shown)


mercury233:
Already done. Thank you.

PR #713, opened 2016-12-19
errata Necrovalley

destdev:
https://ygorganization.com/ocg-text-change-for-necrovalley/
With 20AP Necrovalley gets a small change that makes it affect cards moving themselves out of the grave as well.

### DIFF
--- a/c14799437.lua
+++ b/c14799437.lua
@@ -45,7 +45,6 @@ function c14799437.spfilter(c)
 end
 function c14799437.spcon(e,c)
 	if c==nil then return true end
-	if c:IsHasEffect(EFFECT_NECRO_VALLEY) then return false end
 	local tp=c:GetControler()
 	local sum=0
 	for i=0,4 do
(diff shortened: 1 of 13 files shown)


destdev:
I've just noticed, I missed to adjust summon procedures from the grave to that errata
let me know if I missed any

destdev:
I just got a report telling me, that there is a difference between summon conditions from special summon only monsters (Infernoids+Dystopia+Block Dragon in this case) and summon conditions from normal monsters like Grapha/Machina Fortress/etc, ... thus I revert the above change for special summon only monsters
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=6500&request_locale=ja

I've got a question for another thing I noticed: I can't find any ruling for Machina Fortress at the moment, is there any? I think there are 3 possible ways that can go regarding discarding itself + Necrovalley:
1. Like I did in the last commit, you can't discard itself in that situation. I think it is that one, because you are not allowed to even start performing the summon procedure othervise (see the Grapha ruling in the last answer).
2. You can discard itself, however you don't summon it then. In this case, I don't see a way to prevent the summon from happening, it is currently performed in any case after the procedure enters the operation, no matter what happens there.
3. Necrovalley doesn't affect the card in this case... which is unlikely imo.

PR #721, opened 2016-12-24
fix Traptrix Trap Hole Nightmare

mercury233:
Fix: The spsummon status should reset when the monster leave field and summon again.

Require https://github.com/Fluorohydride/ygopro/issues/1895 fixed to work.


DailyShana:
If the monster spsummoned this turn cost itself to activate its effect, is it able be negated by _Traptrix Trap Hole Nightmare_?
eg. ローンファイア・ブロッサム/Lonefire Blossom and 古聖戴サウラヴィス/Sauravis, the Ancient and Ascended

PR #730, opened 2016-12-30
update Dragon Capture Jar

Hello23-Ygopro:
i want to help wherever i can in whatever capacity .. so ill start with an easy effect
the effect reads in ocg (translated) -

> Reverse: Destroy the "Dragon Tribe / Sealed Vase" existing **on the face side on the field.** When it is destroyed, all the Dragon family monsters that are face-up on the field are attacked.

it also says "on the field" like the english text

right now i cant think of any ocg&tcg effect exists that can sp summon a spell/trap from the spell & trap zone as a monster with custom stats .. all i can think of now is magical hats , but that gets the cards from the deck , and put them facedown to for a short time ..
but if there exists some effect like that the location needs to change , so this can be used as a pre-caution for now

### DIFF
--- a/c55763552.lua
+++ b/c55763552.lua
@@ -14,14 +14,14 @@ function c55763552.filter(c)
 end
 function c55763552.target(e,tp,eg,ep,ev,re,r,rp,chk)
 	if chk==0 then return true end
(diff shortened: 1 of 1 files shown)


Hello23-Ygopro:
apparently LOCATION_ONFIELD doesnt work , so u must use a LOCATION_MZONE+LOCATION_SZONE for eg. , dont know why that is..

PR #732, opened 2016-12-31
Update King's Knight and so on (Conditions of "Cards exists on your field")

Hello23-Ygopro:
The OCG effect reads:

> ①: "Queen's Night" **exists in your field** and can be activated when this card is successfully summoned. Special summon 1 "Jacks Night" from the deck.

Lets say Queen's Knight in your MZone becomes an Equip Card , then its no longer in the MZone , and you wont be able to SpSummon , right ? if this is true , then alot of old cards need to get changed to LOCATION_ONFIELD that have conditions of "exists on your field" , and if their rulings agree

But i think the rulings might be diffrent on King's Knight ( i couldnt find any on Wikia )
since its an old card and when it got released , i assume there were not any effects that turned monsters into Equip Cards that time
i could contact the OCG office to make sure , but i cant speak japanese , sorry

### DIFF
--- a/c64788463.lua
+++ b/c64788463.lua
@@ -15,7 +15,7 @@ function c64788463.cfilter(c)
 	return c:IsFaceup() and c:IsCode(25652259)
 end
 function c64788463.spcon(e,tp,eg,ep,ev,re,r,rp)
-	return Duel.IsExistingMatchingCard(c64788463.cfilter,tp,LOCATION_MZONE,0,1,nil)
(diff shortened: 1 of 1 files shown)


mercury233:
>自分フィールド上に装備カード扱いとなっている「クィーンズ・ナイト」が存在する時に、自分の「キングス・ナイト」の召喚に成功した場合でも、『デッキから「ジャックス・ナイト」１体を特殊召喚』する効果を発動する事ができます。
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=12376&keyword=&tag=-1

yes you are right.

PR #735, opened 2017-01-01
Update Blue-Eyes Alternative White Dragon

Hello23-Ygopro:
OCG effect is : 

> ①: The card name of this card is treated as "blue eyes white dragon" as long as it **exists in the field** · graveyard

Same as : 
https://github.com/Fluorohydride/ygopro-scripts/pull/734 & https://github.com/Fluorohydride/ygopro-scripts/pull/732

[But Not same as](https://github.com/Fluorohydride/ygopro-scripts/blob/master/c35191415.lua#L8) :
> ②: As long as this card is in the **monster zone**, treat the card name as "black magician". 

It looked like , if it does not say the Monster Zone , LOCATION_ONFIELD is used most of the time


Hello23-Ygopro:
Thank You

PR #736, opened 2017-01-01
Update Magician Navigation

Hello23-Ygopro:
OCG effect is : 

> ②: If "Black Magician" **exists in your field**, you can exclude this card from the Graveyard and activate it for 1 Magic / Trap card on the face-up side of the opponent's field

Update for `LOCATION_ONFIELD` , same as : 
-[Dark Magic Attack](https://github.com/Fluorohydride/ygopro-scripts/blob/master/c2314238.lua#L18)
-[Thousand Knives](https://github.com/Fluorohydride/ygopro-scripts/blob/master/c63391643.lua#L19)
etc.

Also , cus of `EVENT_FREE_CHAIN` , added ..
1 `e1:SetHintTiming(0,TIMING_END_PHASE)` , same as : 
-[Eternal Soul ( Special Summon Effect )](https://github.com/Fluorohydride/ygopro-scripts/blob/master/c48680970.lua#L7)
-If I Activate Card Of Demise in my Dark Magician Deck , I cant Special Summon with Magician Navigation so the game asks me in the opponents End Phase

2 `e2:SetHintTiming(0,0x1c0)`, same as : 
-[D/D/D Cursed King Siegfried](https://github.com/Fluorohydride/ygopro-scripts/blob/master/c44852429.lua#L14)
-Any Quick Negate Effect with no Special Condition , etc.


salix5:
1fda4bef1d73247adce9efda80fdebe58e4e5d24
It seems that there is not a clear standard of default timing.
Currently I believe that only some of the following effects need TIMING_END_PHASE in your opponent's turn:
1. once-per-turn EVENT_FREE_CHAIN effect 
If you don't activate it in End Phase, you will waste the chance anyway.
2. Special cards which will be activated most likely in your opponent's End Phase 
e.g. Cyclone or so

For other cards, the system constantly show the hint in EP may be somewhat disturbing.

mercury233:
i wish all special summon effects should have hint in end phase.
for this card, it will miss the change to trigger _Dark Magical Circle_ in opponent's turn.

salix5:
@mercury233 
There are a lot of special summon quick-effects now, and many of them are rarely activated in your opponent's EP, so adding a default hint may interrupt the turn in some unnecessary situation.
In the case of some specific combos, maybe the player should control the timing on his/her own.

Hello23-Ygopro:
@mercury233 , Thank you, thats the more important situation i missed , where the Dark Magician Deck is played because of Dark Magic Circle is so important

@salix5 I can see if to many hints can be distracting , but i think this Ygopro was made to be very close to the Tag Force games , and any official YuGiOh game in general , where they also ask you to activate cards in certain steps
personally i always play with left-click to force Chain Links so I never have a problem , but to have as an option is convenient in general , and for people who forget
but if the people in charge here think End Phase hints is not needed any more , i can respect that

PR #737, opened 2017-01-02
fix SPYRAL Resort

mercury233:
Fix: It don't say "into the Main Deck" like _Caam, Serenity of Gusto_, so it should be able to return extra monster.


nekrozar:
https://github.com/Fluorohydride/ygopro-scripts/pull/689 SPYRAL Resort's pull request already exists.

PR #739, opened 2017-01-02
Update Buster Whelp of the Destruction Swordsman and so on ( Added Commits 01-05-2017 )

Hello23-Ygopro:
The OCG effect is translated as :

> ③: If this card is in the Graveyard and "Buster · Braider" **exists in your field**, you can cast it by discarding 1 "Destroyed Sword" card from your hand. Special summon this card from the Graveyard.

If Buster Blader is an Equip Card , and is your only Buster Blader , u should still be able to SpSummon Buster Whelp

### DIFF
--- a/c21414674.lua
+++ b/c21414674.lua
@@ -15,5 +15,5 @@ end
 function c21414674.spcon(e,c)
 	if c==nil then return true end
 	return Duel.GetLocationCount(c:GetControler(),LOCATION_MZONE)>0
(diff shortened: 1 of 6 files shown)


DailyShana:
for card desc before 9th series (without ①②etc.), I don't know whether the equip cards match the conditions in these effects.

Hello23-Ygopro:
i noticed their OCG effects were outdated .. i guess its better to wait for an updated Effect , or make sure at the FAQ or OCG office before accepting edits on them , Thank you

PR #742, opened 2017-01-04
add missing category & shuffle hand for Allure of Darkness

Hello23-Ygopro:
- Added `CATEGORY_REMOVE` because the effect includes you banishing a DARK monster

- Added `Duel.ShuffleHand(p)`
just like before the discard in [Dark World Dealings](https://github.com/Fluorohydride/ygopro-scripts/blob/master/c74117290.lua#L22) , etc.

You are shuffling your hand cus if your opponent knows about a card in your hand , and they see you banishing / discarding another card of the same name you just drawn with Allure , they still know that the other card is in your hand since you did not shuffle

I dont know how many Ygopro games have the shuffle hand button , but cus it all happens while resolving 1 effect , you dont get a chance to shuffle your hand before you discard / banish , so the script does it for you
# 
**Updated 2017-01-12**
- Added `CATEGORY_TOGRAVE` too
like [Laval the Greater](https://github.com/Fluorohydride/ygopro-scripts/blob/master/c12986807.lua#L9): "When this card is Synchro Summoned: Send 1 card from your hand to the Graveyard."
# 
**Updated 2017-01-18**
- `CATEGORY_HANDES` should be used instead of `CATEGORY_TOGRAVE`
See : #763

### DIFF
--- a/c1475311.lua
+++ b/c1475311.lua
@@ -2,7 +2,7 @@
 function c1475311.initial_effect(c)
 	--Activate
 	local e1=Effect.CreateEffect(c)
-	e1:SetCategory(CATEGORY_DRAW)
+	e1:SetCategory(CATEGORY_DRAW+CATEGORY_REMOVE)
 	e1:SetType(EFFECT_TYPE_ACTIVATE)
 	e1:SetProperty(EFFECT_FLAG_PLAYER_TARGET)
 	e1:SetCode(EVENT_FREE_CHAIN)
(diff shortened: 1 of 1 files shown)


Hello23-Ygopro:
**Updated 2017-01-18*
`CATEGORY_HANDES` might need to be used instead of `CATEGORY_TOGRAVE` : https://github.com/Fluorohydride/ygopro-scripts/pull/763

Someone needs to confirm which is the right Category to use

Issue #599, opened 2016-10-06
Union functions not yet fixed


Hello23-Ygopro:
I just used a-assault core's script for reference and noticed something in its equip function ,

```
function c30012506.eqop(e,tp,eg,ep,ev,re,r,rp)
    local c=e:GetHandler()
    local tc=Duel.GetFirstTarget()
    if not c:IsRelateToEffect(e) or c:IsFacedown() then return end
    if not tc:IsRelateToEffect(e) or not c30012506.filter(tc) then
        Duel.SendtoGrave(c,**REASON_EFFECT**)
        return
    end
    if not Duel.Equip(tp,c,tc,false) then return end
    aux.SetUnionState(c)
end
```

Im no expert , but shouldnt `**REASON_EFFECT**` be `**REASON_RULE**` ?
its sending the monster to the grave because of a game mechanic , being a wrong target
those generally use `REASON_RULE `

salix5:
I agree that it should be REASON_RULE, and maybe all union monsters have the same problem.

Hello23-Ygopro:
@salix5 I dont know if all union monsters have the problem
But im guessing they do since the union update , if that line was copy-pasted , it looks like it was..
Heres some more of the newer union monsters i found with the same lines : 
https://github.com/Fluorohydride/ygopro-scripts/blob/master/c77411244.lua#L72
https://github.com/Fluorohydride/ygopro-scripts/blob/master/c3405259.lua#L72
https://github.com/Fluorohydride/ygopro-scripts/blob/master/c64500000.lua#L73
https://github.com/Fluorohydride/ygopro-scripts/blob/master/c65622692.lua#L73

If you agree it should be REASON_RULE , ill make the few pull requests

Hello23-Ygopro:
Its probably best to fix when links and procuderes and scripts gets compatable first , cus its more important

Thank you for your time

PR #757, opened 2017-01-14
fix Unions equip op

Hello23-Ygopro:
Like i said here - https://github.com/Fluorohydride/ygopro-scripts/issues/599

Before i modify the others , i just wanna make sure that it should be `REASON_RULE` and not `REASON_EFFECT`
i understand it as follows - it sends the equipped monster to the grave as a **rule ( a game mechanic )** because its the wrong target `not c30012506.filter(tc)` or if its not in the same instance as it was when "A-Assault Core" targeted it `not tc:IsRelateToEffect(e)`


Hello23-Ygopro:
Stale branch , revisit on another time

PR #758, opened 2017-01-15
Added Flag for Effects That Activate by Detaching Xyz Materials as cost.

TheOnePharaoh:


c74294676.lua
```diff
@@ -56,10 +57,6 @@ end
 function c74294676.condition2(e,tp,eg,ep,ev,re,r,rp)
 	return Duel.GetCurrentChain()==0
 end
-function c74294676.cost2(e,tp,eg,ep,ev,re,r,rp,chk)
```
TheOnePharaoh: it's the same cost

c77631175.lua
```diff
@@ -14,7 +14,7 @@ function c77631175.initial_effect(c)
 	--damage
 	local e2=Effect.CreateEffect(c)
 	e2:SetDescription(aux.Stringid(77631175,1))
-	e2:SetCategory(CATEGORY_DAMAGE)
+	e2:SetCategory(CATEGORY_ATKCHANGE+CATEGORY_DAMAGE)
```
TheOnePharaoh: Category was missing

mercury233:
what is this flag for?

purerosefallen:
why not a category but a property?

TheOnePharaoh:
@purerosefallen it's a property because it's cost related not effect related.

Hello23-Ygopro:
Thanks , thats a lot of scripts that got updated
but do you mind saying which cards need that flag ? Is it to help some unreleased ocg cards to function properly , or anime/manga/vg/etc, related , or maybe as a pre-caution for future effects ?

c77631175.lua
Hello23-Ygopro: i dont think they care about missing Categorys anymore

PR #761, opened 2017-01-16
Official Number for Zarc in the Summon Procedures

TheOnePharaoh:


TheOnePharaoh:
@VanillaSalt

destdev:
I've already hotfixed that on the Devpro server about 4 hours ago

PR #763, opened 2017-01-17
add missing category for Worm Hope, etc.

Hello23-Ygopro:
OCG effect is :
リバース：このカードが相手モンスターの攻撃によってリバースした場合、デッキからカードを１枚ドローする。また、このカードがフィールド上から墓地へ送られた時、自分は手札を１枚選んで墓地へ送る。
Also, when this card is sent from the top of the field to the Graveyard, you choose 1 card and send it to the Graveyard.

Its an Effect that starts a Chain , so it needs a Category

Same as Laval the Greater ,
effects that includes sending a card from your hand to the Grave : 
https://github.com/Fluorohydride/ygopro-scripts/blob/master/c12986807.lua#L9
And Number 10: Illumiknight : 
https://github.com/Fluorohydride/ygopro-scripts/blob/master/c11411223.lua#L8
etc.

### DIFF
--- a/c11159464.lua
+++ b/c11159464.lua
@@ -10,6 +10,7 @@ function c11159464.initial_effect(c)
 	--discard
 	local e2=Effect.CreateEffect(c)
 	e2:SetDescription(aux.Stringid(11159464,0))
+	e2:SetCategory(CATEGORY_TOGRAVE)
 	e2:SetType(EFFECT_TYPE_SINGLE+EFFECT_TYPE_TRIGGER_F)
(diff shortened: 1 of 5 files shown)


Hello23-Ygopro:
Should CATEGORY_HANDES be used instead of CATEGORY_TOGRAVE , when an effect includes sending a Card from the hand to the gravey , like Gaia, the Mid-Knight Sun ? 
https://github.com/Fluorohydride/ygopro-scripts/blob/master/c40371092.lua#L6

PR #768, opened 2017-01-20
fix Synchro Call

mercury233:
Fix: the missing sync monster limit


Hello23-Ygopro:
Thanks , thats a big fix

PR #771, opened 2017-01-20
update category

mercury233:
for _Ghost Ash & Spring Flowers / 灰流うらら_
more cards need to be updated, but I can't find them...


Hello23-Ygopro:
Try using DataEditorX and press "Reset" and enter key words in the effect field and press "Search"
thats what i did when i looked for specific Category of cards , but its impossable to do in Ygopro Percy cus the databases are split in the live folder , and its better to search in original Chinese/Japanese for OCG rulings

PR #772, opened 2017-01-20
update category for Ghost Ash & Spring Flowers Rulings

Hello23-Ygopro:
https://github.com/Fluorohydride/ygopro-scripts/pull/771

_Void Imagination_ also sends cards from the Main Deck to the Graveyard : 

> If your opponent controls a monster that was Special Summoned from the Extra Deck, and you do not, you can also use up to 6 monsters in your Deck as Fusion Materials.

### DIFF
--- a/c31444249.lua
+++ b/c31444249.lua
@@ -24,7 +24,7 @@ function c31444249.initial_effect(c)
 	c:RegisterEffect(e3)
 	--spsummon
 	local e4=Effect.CreateEffect(c)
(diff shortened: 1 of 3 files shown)


Hello23-Ygopro:
@salix5 These are kinda high priority : 
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=20544&request_locale=ja
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=11500&request_locale=ja

PR #774, opened 2017-01-21
add missing constants

Hello23-Ygopro:
一些較少使用，一些更像使用像聲明屬性，種族，卡名等。
不需要使用這些常量更改現有腳本
它們只是為將來選擇使用常量的十六進制添加的人
摘自https://github.com/Fluorohydride/ygopro/blob/master/strings.conf#L100
從`！系統549和向前開始`
# 

Some are less used , some are more used like declaring Attributes , races , card names , etc.
There is no need to change existing scripts with these Constants
they are just added for anyone who chooses to use Constants over Hex in the future and to make scripting more convenient
taken from https://github.com/Fluorohydride/ygopro/blob/master/strings.conf#L100
starting at `!system 549` and onwards


Hello23-Ygopro:
## Update 2017-01-27
-Added coin selecting constants
-Added reset constants like explained here - https://github.com/Fluorohydride/ygopro/issues/1912

PR #776, opened 2017-01-25
update category

mercury233:


Hello23-Ygopro:
https://github.com/Fluorohydride/ygopro-scripts/pull/772
@mercury233 Thanks , ill just give you the fixes from now on 😂

Edit : 
Could you also post the fix for https://github.com/Fluorohydride/ygopro-scripts/pull/775 , unless im wrong about it needing a damage category

mercury233:
Sorry I forgot you had made that PR...

I made a list of cards(about 500) which may need to add category, but don't have time to check it yet.

Hello23-Ygopro:
No problem
i just used DataEditorX to look for keywords , but cus i dont have all cards and English card text is not official yet for some , i probably missed some
also its better to search for in Japanese anyway , but those 3 are the only ones i could find , and for its Q&A

PR #782, opened 2017-01-27
fix strings

destdev:
e:GetDescription provides "???" on Revival of True Kings and Apocalypse of True Dracos upon the initial activation, there has to be the actual stringid for that

fix the string ids of Humid Winds, that didn't use the 0 string

added opp hints for Shiranui Style Reincarnation, Humid Winds and Tramid Pulse


DailyShana:
`Duel.SelectOption` will give opp hints automatically so it will hint twice if player use the effect when activation of card.
And for all these cards, the two effect can be distinguished easily by observing cost and target.

PR #796, opened 2017-02-06
add tips for Dark Fusion, Vehicroid Connection Zone

Hello23-Ygopro:
Tips for the gained effects , just like these cards have : 
- [Frightfur Sabre-Tooth](https://github.com/Fluorohydride/ygopro-scripts/blob/master/c80889750.lua#L72)
- [Rune-Eyes Pendulum Dragon](https://github.com/Fluorohydride/ygopro-scripts/blob/master/c1516510.lua#L32)
- [Raging Cloudian](https://github.com/Fluorohydride/ygopro-scripts/blob/master/c23639291.lua#L29)

### Vehicroid Connection Zone[23299957]
str0 = 「はカードの効果では破壊されず、。」
str1 = 「効果を無効化されない。」

str0 = "Cannot be destroyed by card effects."
str1 = "Effects cannot be negated."

### Dark Fusion[94820406]
str0 = 「は、このターン相手のカードの効果の対象にならない。」
str0 = "Cannot be targeted by opponent's card effects."


mercury233:
I think _Massivemorph_ don't need tip at most time, because the atk of the target monster is twiced.
The only way to spsummon _Spirit Message_ cards is _Dark Sanctuary_, so it don't need extra tip too.
_Vehicroid Connection Zone_ do need a tip, and the same for **_Dark Fusion_**.

Hello23-Ygopro:
> I think Massivemorph don't need tip at most time, because the atk of the target monster is twiced

I might disagree with that but i can understand too : to overall have less tips as possable , and they should only be added to avoid confusion , and as a reminder if its not obvious enough
For example , like you said when the monsters ATK is doubled , then that is enough information for both players to remember that its affected by *Massivemorph*

> The only way to spsummon Spirit Message cards is Dark Sanctuary, so it don't need extra tip too.

Also if a *Spirit Message* is in the Monster Zone then both players know theyre unaffected so theres no need for hints as a reminder , i understand

> Vehicroid Connection Zone do need a tip, and the same for Dark Fusion.

I thought of *Dark Fusion* too , but cus it resets in the End Phase i didnt think it was needed , but ill add it too, cus one situation i can think of is if someone has a **really** long turn in which they activated *Dark Fusion* then the opponent might forget that its untargatable
So i dont think just cus an effect resets in the End Phase should be enough reason **not** to include tips for them
https://github.com/Hello23-Ygopro/ygopro-scripts/commit/9bdfdc679fc355f7c431be251066607eac1031f6

PR #813, opened 2017-02-26
new card SR04

nekrozar:
http://www.db.yugioh-card.com/yugiohdb/card_search.action?ope=1&sess=1&pid=2201048&rp=99999


Hello23-Ygopro:
@nekrozar i forgot theres also some erratas in that structure deck : 
https://ygorganization.com/kenzanshotfirst/

PR #815, opened 2017-03-02
Fixed Cards that Copy the original name

TheOnePharaoh:
Q.
スターヴ・ヴェノム・フュージョン・ドラゴン
相手が同じモンスターに（2）の効果を与えた場合。 ループを作成できますか？ または、1ターンに1度はまだ有効ですか？
A.
相手フィールドの「スターヴ・ヴェノム・フュージョン・ドラゴン」を対象に、「スターヴ・ヴェノム・フュージョン・ドラゴン」の『②』の効果を発動する事はできません。

Q.
Stave, Venom, Fusion, Dragon
When the opponent gives effect (2) to the same monster. Can I create a loop? Or is it still valid once per turn?
A.
You can not activate the effect of "②" of "Stave, Venom, Fusion, Dragon" for the opponent's field "Stave, Venom, Fusion, Dragon".

Ruling provided by Steeldarkeangel

Note:
Red Supremacy already applies that ruling and it's not a BKSS ruling because
the effect text says `.......this card's name becomes that target's original name, and .......`
the name can't become the target's original name if it already is with that same name
it's the same reason you can't target a monster with 0 ATK and (no negatable effect or if it's effects are already negated) using Clear Wing Fast Dragon
also the same reason you can't declare the current level in the effect of gagaga magician

### DIFF
--- a/c11522979.lua
+++ b/c11522979.lua
@@ -47,14 +47,14 @@ function c11522979.cost(e,tp,eg,ep,ev,re,r,rp,chk)
 	if chk==0 then return e:GetHandler():CheckRemoveOverlayCard(tp,1,REASON_COST) end
 	e:GetHandler():RemoveOverlayCard(tp,1,1,REASON_COST)
 end
-function c11522979.filter(c)
-	return c:IsFaceup() and c:IsType(TYPE_XYZ)
+function c11522979.filter(c,code)
+	return c:IsFaceup() and c:IsType(TYPE_XYZ) and c:GetOriginalCode()~=code
 end
 function c11522979.target(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
 	if chkc then return chkc:IsLocation(LOCATION_MZONE) and chkc:IsControler(1-tp) and c11522979.filter(chkc) end
-	if chk==0 then return Duel.IsExistingTarget(c11522979.filter,tp,0,LOCATION_MZONE,1,nil) end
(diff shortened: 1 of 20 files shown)


EerieCode:
You haven't added the check for the original code in the chk and chkc sections.

TheOnePharaoh:
the code is in the copy filter and the chk checks the whole filter
c41209827.copyfilter(chkc)

EerieCode:
What I mean is that you aren't passing _e:GetHandler():GetOriginalCode()_ in the chk and chkc sections, you're doing it only during target selection.

TheOnePharaoh:
Thanks @EerieCode

TheOnePharaoh:
@mercury233 this has been open for a long time even though it's a simple fix

TheOnePharaoh:
I now just figured out the reason of the ruling 
Because the effect's text says `this card's name becomes that target's original name, and replace this effect with that target's original effects.` which is why it can't target a card with the same original name because you can't change the name to become that monster's original name if both have the same original name.

mercury233:
> Q.
「No.69 紋章神コート・オブ・アームズ」の②の効果は相手フィールド上の「No.69 紋章神コート・オブ・アームズ」を対象に発動できますか？
A.
相手の「No.69 紋章神コート・オブ・アームズ」を対象に自分が「No.69 紋章神コート・オブ・アームズ」の『②』の効果を発動することもできます。

The ruling had been changed

PR #837, opened 2017-04-02
possible way to do link 

ZisIsNotZis:
haven't fully tested yet, but basically tested and everything looks fine


purerosefallen:
But you know, its not permitted to add functions in Group, perhaps you need to change it to aux.

purerosefallen:
In addition, your script format is probably unreadable for others. you need to follow some basic rules eg. c stands for card, g stands for group, and e stands for effect, t stands for table. Otherwise others cannot read your scripts and cannot script other similar cards by your solution.

purerosefallen:
And your link condition forgot to return false at last

ZisIsNotZis:
principally it would be good to write Group stuff in c++ but there's nothing wrong with writing it in lua. Naming optimized, return added

ZisIsNotZis:
issues:
1. It seems that there is no way to get disabled field in lua
2. synchro/xyz checking functions are out of control of lua

purerosefallen:
its possible. you may use Duel.IsPlayerAffectedByEffect to get the effect of EFFECT_DISABLE_FIELD to check which field is disabled

purerosefallen:
but could you give me a solution to change the custom card scripts? there are many of them and i am unwilling to delete or rescript them. this is exactly im worrying about. perhaps i need to learn how to make custom card again...

purerosefallen:
but how about fusion monsters? especially EFFECT_FUSION_MATERIAL which need to be changed for selecting monsters

ZisIsNotZis:
Not sure what do you mean. The general solution is to generate all possible groups based on general rules, them filter out the groups that do not satisfy the constraints required by that specific card. This model can solve any problem, though not as fast as the current solution.

Synchro and XYZ can be done in a really easy way since it do not invoke the change of linked zone. Fusion (with the card "Fusion") might invoke that, but since it do not have to guarantee an available extra zone (according to new master rule), it's even easier. Fusion like "ABC" which have to guarantee an available zone is as hard as link, but we already have the link problem solved using the general solver. As a result, everything is solved.

Hello23-Ygopro:
😭

PR #839, opened 2017-04-03
update moving effects on master rule 4 & The Wicked Avatar

purerosefallen:
these changed cards are related to effect of moving sequence. these should be changed in the master 4 rule because of the ex monster zone

when there are more than one The Wicked Avatar on the field and one of them changed its name by Hero Mask, there will be some problems when calculating attack.


DailyShana:
you should open another pull request for The Wicked Avatar

PR #858, opened 2017-04-29
Merge all the fusion procedures, except AddFusionProcFunFunRep and AddFusionProcCodeFunRep under AddFusionProcMix

edo9300:
I took the procedure made by @VanillaSalt (https://github.com/Fluorohydride/ygopro-scripts/commit/99b75f312a59bd5db2afb4033e820e72d8ff1ba8) and merged the various fusion procedure into one that can use an arbitrary number of materials, still keeping a retro compatibility with the older ones

### DIFF
--- a/utility.lua
+++ b/utility.lua
@@ -541,6 +541,21 @@ end
 function Auxiliary.AddFusionProcCode4(c,code1,code2,code3,code4,sub,insf)
 	Auxiliary.AddFusionProcCode(c,sub,insf,code1,code2,code3,code4)
 end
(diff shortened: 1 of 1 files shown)


edo9300:
Still keeping vanilla's procedure as reference, i changed it to support an arbitrary number of conditions, and i added AddFusionProcFunRep, now AddFusionProcFun2 and AddFusionProcFunRep will call that function

mercury233:
Those procedures in Shaddoll fusion monsters and _Fusion Conscription_ need update too.

PR #859, opened 2017-05-08
"Alligator's Sword Dragon" fix

senpaizuri3:


Hello23-Ygopro:
Sorry but `fixed by senpaizuri is not allowed` in the archives

c3366982.lua
```diff
@@ -1,4 +1,5 @@
 --ドラゴンに乗るワイバーン
+--fixed by senpaizuri
```
TheOnePharaoh: @senpaizuri3 Can you remove this line so the fix would be merged?

PR #860, opened 2017-05-08
Fixed listed materials count

edo9300:
@VanillaSalt now it checks if the code of a monster is already in the materials list


nekrozar:
https://github.com/Fluorohydride/ygopro-scripts/commit/96176da578d17bdaae30c573cc0fdbf6dab0dd22

PR #864, opened 2017-05-19
Fixed "Chaos Scepter Blast" (plus notice for other possible bugs)

EerieCode:
As explained in [this ruling](https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=19209&keyword=&tag=-1), translated [here](https://ygorganization.com/ocg-051817-weekly-rulings-update/) by the Organization, Tokens cannot be banished face-down, so they should be unaffected by this card's effect. The filter I added should be enough to fix the problem: that being said, would it be possible to alter the function Card.IsAbleToRemove() and Card.IsAbleToRemoveAsCost() by adding an optional parameter to check the position the card is going to be banished? By leaving it as POS_FACEUP by default, we'd be able to prevent issues like this and still keep the code short and clear. This isn't the only card with this issue: I can name "Eater of Millions", "PSY-Frame Overload" and "Transmission Gear" right now, there might be even more.

### DIFF
--- a/c15256925.lua
+++ b/c15256925.lua
@@ -28,14 +28,17 @@ end
 function c15256925.condition(e,tp,eg,ep,ev,re,r,rp)
 	return Duel.IsExistingMatchingCard(c15256925.cfilter,tp,LOCATION_MZONE,0,1,nil)
 end
+function c15256925.filter(c)
+  return not c:IsType(TYPE_TOKEN) and c:IsAbleToRemove()
+end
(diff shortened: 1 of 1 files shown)


DailyShana:
fix in https://github.com/Fluorohydride/ygopro-scripts/commit/d85938c27e135ee7ddb2e9fa15eaae795db6a01b#diff-62544683fbc11e5aab0c3bbacaefda3d

PR #866, opened 2017-05-21
Misc Bug Fixes TheOnePharaoh

TheOnePharaoh:
Formula Synchron's Erreta: `Duelist Saga` http://yugioh.wikia.com/wiki/Card_Errata:Formula_Synchron

Assault Dog and Hero Kid: fixed for Blue-eyes Spirit Dragon

Performapal Gatling Ghoul: it `didn't work properly` with cards that change the original ATK of a monster on the field for example shrink or monsters that has ? ATK and determined their original ATK using their effect for example Chimeratech Overdragon.
`and if you do, inflict damage to your opponent equal to the original ATK the destroyed monster had on the field.`

Added pendulum limitation from the master rulebook: ペンデュラムモンスターはモンスターカードではありますが、ペンデュラムゾーンに置く際には、魔法カードとして発動し、その処理によって置かれる事になります。
　（ペンデュラムモンスターをペンデュラムゾーンにセットする事はできません。）
Pendulum monster is a monster card , but when putting it in the pendulum zone it will be activated as a magic card and will be placed by that process.
`(You can not set a pendulum monster in the pendulum zone .)`


mercury233:
Sorry for the long wait.

Is there any card that can turn cards in S&T zone to face-down? BTW, I think it is better to do this check in https://github.com/Fluorohydride/ygopro-core/blob/261de912bae65b446bd7734b6ba060d6c6d18c31/card.cpp#L3313 .

PR #874, opened 2017-06-04
Added mg and sg to the fusion filters

edo9300:
Updated dante, pilgrim of the burning abyss and pair cycroid


mercury233:
could be slow when mg is large..

salix5 (commented):
@mercury233 @edo9300 
utility.lua
Line 649:
It seems that `mg` is undefined here.
Should we revert this?

PR #876, opened 2017-06-11
firewall dragon's effect can be activated in damage step

zhykzhykzhyk:
When an effect can be both activated when it has been destroyed by battle or effect, this effect can be activated in battle step even if the card is destroyed by effect, like "Torrential Reborn".


mercury233:
Effects at `EVENT_BATTLE_DESTROYED` don't need `EFFECT_FLAG_DAMAGE_STEP` now.

zhykzhykzhyk:
But EVENT_TO_GRAVEYARD needs it

PR #878, opened 2017-06-13
Added shortcut for self-banish cost

EerieCode:


mercury233:
`aux.bfgcost(e,tp,eg,ep,ev,re,r,rp,0)` is not shorter than `e:GetHandler():IsAbleToRemoveAsCost()` , so i don't think we should change them

PR #888, opened 2017-06-28
Added function Card.IsOnSameColumn

EerieCode:
This new function is a shortcut for the filter that checks whether two cards are in the same column. Given the new importance of positioning in MR4, and the fact that 2 cards in Circuit Break so far required this check, I figured it'd be better to have a shortcut in the utilities, similar to the ones for self-banishing cost and whether a card was sent to the Graveyard during that turn. The function includes an optional parameter to indicate whether the Extra Monster Zone should be checked (required for Blasting Fuse): by default, the Extra Monster Zone will always be checked.

### DIFF
--- a/utility.lua
+++ b/utility.lua
@@ -1369,3 +1369,26 @@ function Auxiliary.bfgcost(e,tp,eg,ep,ev,re,r,rp,chk)
 	if chk==0 then return e:GetHandler():IsAbleToRemoveAsCost() end
 	Duel.Remove(e:GetHandler(),POS_FACEUP,REASON_COST)
 end
+--Checks whether 2 cards are on the same column
+--skip_ex is optional, indicates whether the Extra Monster Zone should be ignored (used in Blasting Fuse)
+function Card.IsOnSameColumn(c1,c2,skip_ex)
+	if not c1 or not c1:IsOnField() or not c2 or not c2:IsOnField() then return false end
+	if c1==c2 then return false end
+	local s1=c1:GetSequence()
+	local s2=c2:GetSequence()
+	if (c1:IsLocation(LOCATION_SZONE) and s1>=5)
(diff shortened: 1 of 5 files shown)


nekrozar:
In Data Base, _Blasting Fuse_ don't destroy.
> https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=4&cid=6910
> ■発動した「爆導索」自身は破壊されません。

purerosefallen:
I think this function should be in ocgcore or begin with aux. It is very strange to put any functions which belong to Card in utility.lua

mercury233:
+1 for @purerosefallen

EerieCode:
@nekrozar It seems like the original script was bugged too, then, since "Blasting Fuse" was also included in the group. I'll upload a patch once I return home, I can't change the files ATM. Thanks!

@purerosefallen @mercury233 Unfortunately, I'm not familiar enough with the core to add it there. It's like when I proposed Card.IsSummonType months ago: it's a function that should logically belong to Card, rather than Auxiliary, but I can't put it directly in the core.

edo9300:
@mercury233 well, Card and Group functions, have the advantage of being called by writing `c:Function` instead of `aux.Function(c)`

EerieCode:
Ok, "Blasting Fuse" should be fine now. As for the Card in the utilities issue, may I recommend approving it anyway, at least temporarily? If you decide to add it to the core, it can easily be removed later, but at least this way the new shortcut will be avaiable already.

utility.lua
```diff
@@ -1383,7 +1383,7 @@ function Card.IsOnSameColumn(c1,c2,skip_ex)
 			return s2==s1
 		else
 			return s2==s1 or (s1==1 and s2==5) or (s1==3 and s2==6)
-		else
+		end
```
mercury233: @EerieCode Please don't forget to do some basic tests next time.

PR #891, opened 2017-07-05
Update utility.lua

Yumekaki:


purerosefallen:
i think you need to update some card scripts with this function

mercury233:
see https://github.com/Fluorohydride/ygopro-pre-script/issues/133

PR #893, opened 2017-07-14
fix Firewall Dragon

mercury233:
The `EFFECT_FLAG_CVAL_CHECK` is not needed here.

But it may has issue like _Metalfoes Counter_(https://github.com/Fluorohydride/ygopro/issues/1917), so it should be updated.

@VanillaSalt I'm not sure about `Duel.RaiseEvent`, is this way correct or should we use `Duel.RaiseSingleEvent`?


VanillaSalt:
`RaiseSingleEvent` is better

PR #896, opened 2017-07-16
fix

mercury233:
https://github.com/Fluorohydride/ygopro-scripts/commit/df33ef84376982b2aa8d5a5fbb23f6209a5f6d72#diff-e2dcb2b3468cae90b8ab3113481f4879L627

@VanillaSalt Is it intentional to remove this dialog?


VanillaSalt:
Yes. I think it's inconvenient that many clicks are needed.

mercury233:
@VanillaSalt But the user hardly can know that he can use the finish button or right mouse button to finish it, so I think this dialog is necessary.

mercury233:
@VanillaSalt Sorry I'll have to add the dialog back before we get a better way to let the users know the finish button or right mouse button can be used.

PR #909, opened 2017-08-03
Fixed "Black Garden"

EerieCode:
https://github.com/Fluorohydride/ygopro/issues/2026

Changing the summoner when the token is to be summoned on the opponent's side of the field should be enough to solve the issue.


raphael-costa:
This is still incorrect, the owner of Black Garden is the player summoning the tokens, this commit would allow players to activate Black Garden, activate Maxx "C" and draw off their own summons.

PR #912, opened 2017-08-08
update Link In-Flyer

nekrozar:
https://github.com/Fluorohydride/ygopro-core/pull/109

### DIFF
--- a/c65100616.lua
+++ b/c65100616.lua
@@ -18,5 +18,5 @@ function c65100616.spcon(e,c)
 	return Duel.GetLocationCount(tp,LOCATION_MZONE,tp,LOCATION_REASON_TOFIELD,zone)>0
 end
 function c65100616.spval(e,c)
-	return Duel.GetLinkedZone(c:GetControler())*0x10000
+	return bit.lshift(Duel.GetLinkedZone(c:GetControler()),16)
 end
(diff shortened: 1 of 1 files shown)


purerosefallen:
But i think the value of spsummon procs are the summon type. Eg. SUMMON_TYPE_XYZ
So how to deal with that?

nekrozar:
`SUMMON_TYPE` is used 0x1000000 or higher.

purerosefallen:
However, some cards use a number below that as a summon type. eg. Gladiator Beasts

purerosefallen:
Maybe you need to use 0xff as the zones and the number above 0x100 as the summon type. So you would have to change the Gladiator Beast scripts

nekrozar:
https://github.com/Fluorohydride/ygopro-core/blob/master/operations.cpp#L2460
`EFFECT_SPSUMMON_PROC` don't read from 0x10000 to 0xff0000.

purerosefallen:
Well, thats ok.
And could you update EFFECT_SPSUMMON_PROC_G as well?

purerosefallen:
And.. Using *0x1000 seems a little bit ugly. Could you change it into bit.lshift?

nekrozar:
>And could you update EFFECT_SPSUMMON_PROC_G as well?

`EFFECT_SPSUMMON_PROC_G` is used Pendulum Summon only.
So I think unnecessary.

>Could you change it into bit.lshift?

https://github.com/Fluorohydride/ygopro-scripts/pull/912/commits/633eca893ca9125729acfd8f82c4582a979d5c90 Updated.

PR #932, opened 2017-09-20
fix Phantasm Spiral Grip and so on

destdev:
added missing checks to several equip cards to make sure, that the equipped monster is controlled by the same player who controls the equip card
effects of equip cards that activate upon destroying an opponents monster by battle should not activate if the player who controls the monster is not the same player who controls the equip card


mercury233:
didn't consider _Ancient Lamp_

PR #960, opened 2017-10-20
update gagaga magician

AndreOliveiraMendes:
update to use Announce Level


nekrozar:
Why you change _Gagaga Magician_ only?

AndreOliveiraMendes:
idk

PR #964, opened 2017-10-21
Zone moving cards update

edo9300:


purerosefallen:
Please change all `bit` functions into the bit operators.

PR #966, opened 2017-10-26
Updated guy dance

edo9300:
The opponent should be able to know the zone selected be the player who activates the card, with that workaround he'll be forced to select the same zone selected by the player as confirm.

### DIFF
--- a/c50696588.lua
+++ b/c50696588.lua
@@ -14,12 +14,12 @@ function c50696588.cost(e,tp,eg,ep,ev,re,r,rp,chk)
 end
 function c50696588.target(e,tp,eg,ep,ev,re,r,rp,chk)
 	if chk==0 then return Duel.GetLocationCount(1-tp,LOCATION_MZONE,PLAYER_NONE,0)>0 end
-	local flag=Duel.SelectDisableField(tp,1,0,LOCATION_MZONE,0)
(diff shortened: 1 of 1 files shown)


c50696588.lua
```diff
@@ -14,12 +14,12 @@ function c50696588.cost(e,tp,eg,ep,ev,re,r,rp,chk)
 end
 function c50696588.target(e,tp,eg,ep,ev,re,r,rp,chk)
 	if chk==0 then return Duel.GetLocationCount(1-tp,LOCATION_MZONE,PLAYER_NONE,0)>0 end
-	local flag=Duel.SelectDisableField(tp,1,0,LOCATION_MZONE,0)
+	local flag=Duel.SelectDisableField(1-tp,1,LOCATION_MZONE,0,bit.bnot(bit.lshift(Duel.SelectDisableField(tp,1,0,LOCATION_MZONE,0))))
```
DailyShana: do not change the value of flag

edo9300: In that way, the returned flag will be the one refereed to the opponent, so it will avoid doing the lshift later

DailyShana: The value of `EFFECT_MUST_USE_MZONE' is the zone flag in view of the handler player of the effect.

PR #1013, opened 2018-03-30
fix 24010609 reported by YGOCore Unite

purerosefallen:
qq 20180330132308

### DIFF
--- a/c24010609.lua
+++ b/c24010609.lua
@@ -130,5 +130,4 @@ function c24010609.setop(e,tp,eg,ep,ev,re,r,rp)
 		tc:RegisterEffect(e1)
 		tc=tg:GetNext()
 	end
-	Duel.ShuffleSetCard(tg)
 end
--- a/c24010609.lua
+++ b/c24010609.lua
@@ -69,7 +69,7 @@ function c24010609.chainlm(e,rp,tp)
 end
 function c24010609.regop(e,tp,eg,ep,ev,re,r,rp)
 	local c=e:GetHandler()
-	if re:GetHandler():IsSetCard(0x115) and re:IsActiveType(TYPE_SPELL) and rp==tp and re:IsHasType(EFFECT_TYPE_ACTIVATE) then
+	if re:GetHandler():IsSetCard(0x115) and re:IsActiveType(TYPE_SPELL) and rp==tp and re:IsHasType(EFFECT_TYPE_ACTIVATE) and re:GetHandler()~=c then
 		local flag=c:GetFlagEffectLabel(24010609)
 		if flag then
 			c:SetFlagEffectLabel(24010609,flag+1)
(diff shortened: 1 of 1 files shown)


Sonic714:
Updated my commit based on yours.
It is suggested that all SSET operations should be added into one command or Flying Fortress SKY FIRE could only destroy one of the set cards instead of all.
According to the test yesterday, variable ct seemed not to accumulate when the card itself was activated. So checking re:GetHandler()~=c might be unnecessary.
Removing confirmation is a quick way to keep position infomation private. This effect, however, is untargetable, and will make opponent unable to learn which cards leave GY if he does not keep the cards in GY in mind.

purerosefallen:
however, a core change will be needed to support setting multiple cards at the same time, which may only be in next version.
but in some places, especially YGOCore Unite, players kept on arguing on this card bug. so we need to find out a quick way to solve the problem.

Sonic714:
It is useless to try ShuffleSetCard cause it cannot shuffle set spell or trap cards, as well as ConfirmCards.

Sonic714:
SSET param could be Group, and therefore it is possible to be implemented for this card. You can refer to my commit. But cards such as Set Rotation and The Shallow Grave remain unsolved cause setting cards onto both field is unimplemented for current core.
There is a buggy approach to erase the opponent's notes of the set cards, that is, to go the long way. Send them to deck, shuffle, and then set them from deck. But it might conflict with Oracle of Zefra's syncro effect.

mercury233:
I think we should fix the core instead. BTW, shuffle isn't right way since the sequence is important too.

purerosefallen:
I tend to close this pr now, for this pr and #1014  made into a large argument in YGOCore Unite, but useless and not solving the problem.
What I want to say is, it's not correct to force others to THANK to your contribution. It's also not suitable to say anything bad for not putting you on the credits list. Every contributions, every commits and every pull requests in this community are voluntary.
I think it's a little bit my fault. As my couple, Sekka, the Head Judge and the Head of Organizational Department of YGOCore Unite, always tend to put my name on the thank list, which I had started to take for granted. But it was only then that I realized It may run into an argument for some unexpected reasons.
That's all what I want to say. And it's too late now. I'm going to bed. Good night.

Sonic714:
I do apologize for what I have done. And an accommodation between YGOCore League and me has been reached now.
The whole thing started from I spending two days in handling the bugs on my own. But I was ignored and even rumored to make no contribution to fixing the bugs by certain individuals. Thus, in a moment of anger I mistakenly felt my fruits of labors were robbed and faced with unfair treatment, and thought YGOCore League was to blame. This was a serious fault of mine. Nevertheless, I did NOT OBLIGE anyone.
YGOCore League is a voluntary organization as well as a close friend for maintaining a good cooperation partnership with me for more than three years. I ought to choose a more proper way to communicate instead of criticize. Above all, all of these happened when I volunteered to help YGOCore League, so it was highly unreasonable for me to make such acts.
Please allow my thanks and apologies as a commentator and developer. I will do my duty seriously from now on and prevent further mistakes.

PR #1014, opened 2018-03-30
Fixed c24010609 Based on #1013

Sonic714:
Fixed the issue where an additional field spell card could be set from GY.
Combined SSET operation to avoid bugs with Flying Fortress SKY FIRE.
Improved SSET experience for users.


Sonic714:
Besides, aux.Stringid(24010609,2) needs changing to "是否盖放场地魔法卡？".

purerosefallen:
see my commits below

PR #968, opened 2017-10-27
Add Auxiliary.CheckGroup & Auxiliary.SelectGroup using Group.SelectUnselect

purerosefallen:
Auxiliary.CheckGroup(group,group_filter,group_selected,min,max,...)
Auxiliary.SelectGroup(tp,description_text,cancelable,group,group_filter,group_selected,min,max,...)
Also updated Xyz/Link/Fusion/Ritual procs using this function
https://github.com/Fluorohydride/ygopro-core/pull/87 is needed

PR #1015, opened 2018-03-30
update link procedure to use Group.SelectUnselect

mercury233:
https://github.com/Fluorohydride/ygopro/pull/1963


purerosefallen:
take a look at @ #968 which may be a universal solution

edo9300:
@purerosefallen i think for the procedures, it would be better to avoid using that kind of function, since it could be required some changes due to some cards/effects that need a major control over the loop

PR #1019, opened 2018-04-20
Update constant.lua

Yumekaki:


purerosefallen:
I think SEQ_MZONE_MAIN and SEQ_MZONE_EXTRA and SEQ_MZONE_ALL are needed. So are stuff of SEQ_SZONE and the opponent's

PR #1020, opened 2018-04-20
fix Metamorphosed Insect Queen

destdev:
it should use the same token as the original Insect Queen


mercury233:
But Ojamas use different token id, so this card should use different token too.

destdev:
Those tokens are all different because of the different coloured ojamas on the card artwork ... I guess?
I don't see the reason to make any difference here, it's one and the same token with one and the same artwork ... unless I'm missing something?

mercury233:
In case of konami release a different token in future, we should use different token id.

PR #1022, opened 2018-04-21
Solution 1: without mass altering

Aephiex:
See https://github.com/Fluorohydride/ygopro-core/pull/148 for details.


purerosefallen:
I don't think we should overwrite the functions in Duel lib in utility.

PR #1027, opened 2018-04-25
Add comment to link markers

Aephiex:


purerosefallen:
What are those commemts for?

Aephiex:
Does what it seems.

PR #1034, opened 2018-04-27
misc update

Aephiex:
1. more "ExceptThisCard"
2. some special summon procedures


purerosefallen:
Besides, i don't think every material group should be with LevelFreeFilter, for dealing with alter xyz procedures.

PR #1035, opened 2018-04-28
Less annoying tricksters

Aephiex:
Make tricksters' "Duel.Hint" once per turn so that less annoying.


mercury233:
I'll make the user can turn off the hint in client

purerosefallen:
I don't think it should be turned off on the 2nd burn per turn. It may make players unaware of why their LP are lost.
Anyway, to make an option to turn off the hint may be fine.

PR #1040, opened 2018-05-01
Fix Gagaga Head

Aephiex:
Issue:
1. I summon "Gagaga Head" then use it's effect to special summon two Gagaga monsters from my graveyard. For the rest of the turn, I can't special summon except by Xyz summon using only Gagaga monsters as Xyz material.
2. I destroy a "Number 53: Heart-eartH" without overlay cards on my side of the field with "Raigeki Break".
3. I activate the effect of "Number 53: Heart-eartH" and special summon "Number 92: Heart-eartH Dragon" to the field.

Oops, what the hell?

After the patch:
It worked properly.


Aephiex:
Current logic:
1. Until the end of the turn, you cannot special summon unless by xyz summon.
2. Until the end of the turn, cards in your 0x7f and your opponent's 0x7f cannot be used as xyz material of xyz monsters you control, as long as that card is not a Gagaga.
3. Until the end of the turn, you gain a flag that prevents you from xyz summoning without materials (custom card situation).

PR #1047, opened 2018-05-09
Smarter Synchro Material Filter

Aephiex:
1. Make it so that c1 can be taken into consideration while selecting c2. The same is true while selecting c3 and other synchro materials beyond those.
2. Removed the confusing "80896940" and use smarter synchro material filter for majestic instead.

### DIFF
--- a/c67030233.lua
+++ b/c67030233.lua
@@ -48,8 +48,8 @@ end
 function c67030233.mfilter1(c)
 	return c:IsCode(21159309)
 end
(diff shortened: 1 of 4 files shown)


purerosefallen:
I think it's better to buy a better CPU.

Aephiex:
Say that to your server.

AndreOliveiraMendes:
i think the less heavy possible you can make it better, because opmitization is a thing

PR #1049, opened 2018-05-11
fix missing must material check

Aephiex:


purerosefallen:
EFFECT_MUST_BE_FMATERIAL was checked in fusion procedures, rather than in every cards.

Aephiex:
read before comment.

PR #1052, opened 2018-05-13
fix Metaphys Ragnarok and so on

nekrozar:
Fix this: If player use "banish the top N cards of your Deck" effect, player's deck shuffle.


purerosefallen:
in the 1st script, it should only disable shuffle check when there are banishable cards.

PR #1053, opened 2018-05-15
update dice effects

destdev:
update card effects for the new card No.67 パラダイスマッシャー in CP18 that have problems if the dice results are 7


purerosefallen:
I think we should wait for Konami's answer about that. Perhaps some old card texts are to be updated.

destdev:
I noticed myself, due to mercurys most recent commit most of it isn't really too necassary anymore. Only the current text of 32015116 has problems with the current script, so yes, let's just wait for an answer or a text update, I don't think there is too big of a rush for that one^^

destdev:
Because I wasn't fully awake when you asked me to do this and just did changes to all the files I did previously without wasting a thought about if it is still necessary?
Meh, I'll just revert those^^

Issue #947, opened 2017-10-10
Update scripts to use Duel.AnnounceLevel and Card.IsLevel/Rank/Link


mercury233:
Since those functions are added, scripts can be updated. But I don't have time recently.

@AndreOliveiraMendes @edo9300 if you want to do this, PR is welcomed.

AndreOliveiraMendes:
Can it be little by little then?

PR #1055, opened 2018-05-17
update level/rank/link functions

destdev:
adress the issue https://github.com/Fluorohydride/ygopro-scripts/issues/947
this commit updates all those GetLevel/Rank/Link==xx to IsLevel/Rank/Link
also, updated some GetLevel>=xx or <=xx functions to use IsLevelAbove/Below instead
I counldn't find any files that would need Duel.AnnounceLevel updates, all those seem to have been done already?


purerosefallen:
How about the scripts in ygopro-pre-script?

destdev:
I'll do those tomorrow.

PR #1056, opened 2018-05-19
update standard reset values

destdev:
I've recently found the standards for for the most common reset vanues again. I think they where introduced months ago (some time last year?) after a discussion that those might be usefull. Well, I thought and still think so too, however since nobody ever used them and everyone just copies the whole standard lines from other scripts without ever taking a look in the constant.lua they became forgotten very fast ...
I'd say, let's change that and here is the result, every single 0x1fe0000 and 0x1ff0000 replaced with RESETS_STANDARD and RESETS_STANDARD_DISABLE


purerosefallen:
In fact there are many resets. Such as `0x1fc0000` and `0xfe0000` and `0xfc0000` and even `0x47e0000`. Adding such constants will make the scripts messy.
Also `TYPES_EXTRA` is still under consideration.

mercury233:
it can be `RESETS_STANDARD+RESET_XXX` or `RESETS_STANDARD-RESET_XXX`, but i'm too lazy to do it

destdev:
I didn't add those constants, they were already there in the constant file for ... no idea how long exactly, should be at least a year I think (after looking: 10th February 2017). If I remember correctly, they were added after a somewhat lengthy discussion. I can't really say much to that, I don't think I participated there ... if I remember correctly the result was to make the constants in order to avoid confusion of what these values in all those reset lines mean (and I thought replace them eventually?).
Anyway, I forgot about those until now.
Soooo ... all in all is that a general no then? In that case I'm quite confused why they even exist if they aren't used at all^^

purerosefallen:
Then why not adding a constant RESETS_STANDARD_NOTURNSET and RESETS_STANDARD_NOTOFIELD

mercury233:
it is not easier to use than `RESETS_STANDARD+XXX`

purerosefallen:
So why RESETS_STANDARD_DISABLE was specially treated?

mercury233:
it is common enough.

destdev:
I found that in 404 instances, yes that is common enough. The other values ... shouldn't be too common...
I'll quickly search how many there are:
0x1fc0000 117 matches ... more than I thought
0xfe0000 51 matches
0xfc0000 2 matches 
0x47e0000 80 matches
with 250 matches around all 4 values, not sure if that's enough to get an extra constant for any of those, maybe for 0x1fc0000, but else I'd say no. I'll simply update it with Standard+-xxx for now, maybe

mercury233:
BTW, https://github.com/Fluorohydride/ygopro-scripts/pull/318

nekrozar:
Maybe you forget `0x7e0000`, `0xff0000`, `0x17a0000`, `0x1ec0000`, `0x3fe0000`, `0x4760000`, `0x47c0000` and `0x57a0000`.

PR #1057, opened 2018-05-19
update Card.IsLevel/Rank/Link

nekrozar:
https://github.com/Fluorohydride/ygopro-core/pull/163


purerosefallen:
I think we should add `Card.IsLevelBetween`.

PR #1061, opened 2018-06-03
Update c15661378.lua

Aephiex:
Fusion materials used for this can be special summoned with "De-Fusion".
Fixed it so that Pendulum Fusion can fusion summon this card using cards in your pzone for material.

Already tested myself.

Known issues:
1. Any custom card that may fusion summon this card from graveyard or banished will ignore it where only monsters you control or in your hand can be selected.
2. Any custom card that may return this card to extra and special summon the materials later this turn will not be able to special summon the materials.

### DIFF
--- a/c15661378.lua
+++ b/c15661378.lua
@@ -3,6 +3,14 @@ function c15661378.initial_effect(c)
 	--fusion summon
 	c:EnableReviveLimit()
 	aux.AddFusionProcFunRep(c,c15661378.ffilter,3,false)
+	--ffilter swap
+	local e0=Effect.CreateEffect(c)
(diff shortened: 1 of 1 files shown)


Aephiex:
@mercury233

MonokuroInzanaito:
I don't think it makes any sense.
The flag effect will disappear after it returns to the extra deck.
See my PR.

Aephiex:
Let me explain how it works.
An effect that says "EVENT_ADJUST" will NOT excute while another effect is still excuting. If you return that card to your extra with De-Fusion, this effect will just keep waiting until you have done special summoning it's materials.

Aephiex:
In addition, I have tested this in my custom puzzle and it worked just fine.

PR #1064, opened 2018-06-05
Added second material filter for Trishula, the Ice Prison Dragon

edo9300:
Wiith this change, if you set a function as a value in the fusion
procedure effect returned by the auxiliary function, that function will
be called only when performing a fusion summon

### DIFF
--- a/c15661378.lua
+++ b/c15661378.lua
@@ -2,7 +2,8 @@
 function c15661378.initial_effect(c)
 	--fusion summon
 	c:EnableReviveLimit()
-	aux.AddFusionProcFunRep(c,c15661378.ffilter,3,false)
+	local eff=aux.AddFusionProcFunRep(c,c15661378.ffilter,3,false)
+	eff[1]:SetValue(c15661378.matfilter)
 	--spsummon condition
 	local e1=Effect.CreateEffect(c)
 	e1:SetType(EFFECT_TYPE_SINGLE)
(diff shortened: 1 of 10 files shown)


purerosefallen:
Why not using the current 0x100 instead?
In addition, every value function should begin with e as the 1st param, or it will run into redtext.
Also, having a 0x10000 will disturb the 0x100 detection, so place &0xf at the 0x100 check.
But why you return the effect in a table?

edo9300:
@purerosefallen so, 1st question, I'm not using the 0x100 because when you're using it it also checks for fusion substitutes, and this shouldn't happen when only checking materials. For the value, in this case there's no need to put e as parameter, since that value is only used to store another filter, its not used anywhere else in the core, so it won't cause problems, for the & 0xf you're right, I forgot to put it when I ported the functions from my procedure. For the table, I'm returning the effect as a table in case of an hypothetical change in the procedure that could lead to return multiple values, in this way there shouldn't be issues.

Aephiex:
It's a good idea, but it would be better if it's more intuitive that in which case the value function is ignored. For example: `fusc:CheckFusionMaterial(mg,c,PLAYER_NONE,true)`


>purerosefallen:
>In addition, every value function should begin with e as the 1st param, or it will run into redtext.

The function is never called anywhere else than "Auxiliary.FConditionFilterMix" so this shouldn't matter.

edo9300:
@Aephiex well, that's the only way to pass parameters in the fusion procedure without using weird workarounds/changing the core, and I have to said we're lucky too that it's an int value, allowing more checks. Maybe it could be added as a constant, but I don't really like the idea

purerosefallen:
In fact, the value function will be called in every adjust timing, in which the effect itself will be passed into the 1st param.
So having a non-effect param at the 1st place in a function value will result into redtext.

edo9300:
It won't be called, since the core never calls a value from that effect, same in adjust steps.

PR #1065, opened 2018-06-13
update utility

nekrozar:
Update this: Use `pairs`

### DIFF
--- a/utility.lua
+++ b/utility.lua
@@ -1835,10 +1835,12 @@ function Auxiliary.GetLinkCount(c)
 	else return 1 end
 end
 function Auxiliary.LCheckOtherMaterial(c,mg,lc)
-	local le=c:IsHasEffect(EFFECT_EXTRA_LINK_MATERIAL)
-	if not le then return true end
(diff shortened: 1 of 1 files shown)


utility.lua
```diff
@@ -1835,10 +1835,12 @@ function Auxiliary.GetLinkCount(c)
 	else return 1 end
 end
 function Auxiliary.LCheckOtherMaterial(c,mg,lc)
-	local le=c:IsHasEffect(EFFECT_EXTRA_LINK_MATERIAL)
-	if not le then return true end
-	local f=le:GetValue()
-	return f(nil,lc,mg)
+	local le={c:IsHasEffect(EFFECT_EXTRA_LINK_MATERIAL)}
```
DailyShana: if `le` is nil, it should return true here.

nekrozar: https://github.com/Fluorohydride/ygopro-scripts/pull/1065/commits/9f594b9441cd171feeff01de5f44322a4ac57dce fixed

PR #1067, opened 2018-06-22
update NecroValley

DailyShana:
implement the idea of https://github.com/Fluorohydride/ygopro/issues/2022#issuecomment-397996127
only update 2 cards as sample yet


purerosefallen:
Too much work to do. It will destroy the custom card communities.

Aephiex:
I have a plan. Actually, I have three.
1. We can untarget a card when it's sent to another zone, before the effect itself to resolve, so that an effect won't be countered because of one of it's targets is in a graveyard as it resolves. _(Core Edit Required.)_
2. We can mark an effect as "likely to be countered by necrovally" as it activates if if targets one or more cards in a graveyard. Only those will be countered.
3. `Duel.SetOperationInfo(0,CATEGORY_REMOVE,*g*,gc,0,0)` → `Duel.SetOperationInfo(0,CATEGORY_REMOVE,*nil*,gc,0,0)`

PR #1068, opened 2018-06-24
fix & update Cocoon of Ultra Evolution ad so on

nekrozar:
Update: this: Use `Duel.GetMZoneCount`.
Fix this: https://github.com/Fluorohydride/ygopro-core/pull/169


purerosefallen:
Please don't fix multiple things in a single PR.
Fix the 2nd thing by opening another PR instead.

PR #1077, opened 2018-07-17
fix Revendread Evolution

destdev:
missing substraction from the rc level in mfilterf, which (rarely) can possibly lead to an error in the operation if the selected card on the field leaves no (or not enough) levels for any other possible materials

### DIFF
--- a/c7986397.lua
+++ b/c7986397.lua
@@ -31,8 +31,9 @@ end
 function c7986397.mfilterf(c,tp,mg,dg,rc)
 	if c:IsControler(tp) and c:IsLocation(LOCATION_MZONE) and c:GetSequence()<5 then
 		Duel.SetSelectedCard(c)
-		return mg:CheckWithSumEqual(Card.GetRitualLevel,rc:GetLevel(),0,99,rc)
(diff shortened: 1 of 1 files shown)


DailyShana:
I think you misunderstand `Duel.SetSelectedCard(c)`, it set `c` as selected cards in `mg` when calling `mg:CheckWithSumEqual(...)` or `mg:SelectWithSumEqual(...)` but not remove it from `mg`.
The operation of this card should be rewrited using `SelectWithSumEqual`.

destdev:
You are correct, I did not take that into account. I traced back a lonely error we had in our server log and thought it should be according to wrong calculations within that part I edited. Unfortunately I can't really get any details of what exactly happened during that duel, however I was able to narrow the issue down to some wrong calculations while having a full field.
I found an error in my original approach and I think I found the actual faulty calculation now. The dg group is merged with mg in the operation, while not merged with it in the checks. The only place that doesn't take that into account is the mfilterf, thus rarely allowing cards to be selected on that full field that would need 2 or more materials from the deck in order to work properly. I changed that now. I didn't notice anything else that could be wrong, feel free to check again.

PR #1086, opened 2018-08-06
fix for Ancient Lamp

AndreOliveiraMendes:
was not checking if card can be attack target.

### DIFF
--- a/c54912977.lua
+++ b/c54912977.lua
@@ -26,9 +26,9 @@ function c54912977.condition(e,tp,eg,ep,ev,re,r,rp)
 end
 function c54912977.target(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
 	if chkc then return chkc:IsControler(tp) and chkc:IsLocation(LOCATION_MZONE) end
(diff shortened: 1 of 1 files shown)


c54912977.lua
```diff
@@ -26,9 +26,9 @@ function c54912977.condition(e,tp,eg,ep,ev,re,r,rp)
 end
 function c54912977.target(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
 	if chkc then return chkc:IsControler(tp) and chkc:IsLocation(LOCATION_MZONE) end
-	if chk==0 then return true end
+	if chk==0 then return Duel.IsExistingTarget(Card.IsCanBeBattleTarget,tp,0,LOCATION_MZONE,1,Duel.GetAttacker(),Duel.GetAttacker()) end
```
DailyShana: It's a mandatory effect so this line is unnecessary.

Gideon9212: . Before damage calculation, if this card is being attacked by an opponent's monster, and was face-down at the start of the Damage Step: You **can** target 1 monster your opponent controls, except the attacking monster; the attacking monster attacks it instead, and you proceed to damage calculation.

Sorry last I checked, both TCG and OCG text wise, it says **you can*. It is a effect the user can decline to use.

PR #1096, opened 2018-10-05
update EFFECT_NONTUNER

nekrozar:
https://github.com/Fluorohydride/ygopro-core/pull/190

### DIFF
--- a/c10755984.lua
+++ b/c10755984.lua
@@ -1,7 +1,7 @@
 --ダイガスタ・イグルス
 function c10755984.initial_effect(c)
 	--synchro summon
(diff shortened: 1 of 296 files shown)


DailyShana:
changing the interface should be avoid.
we can update `aux.NonTuner` as follows:
```
function Auxiliary.NonTuner(f,...)
	--
	return	function(target,syncard)
				return target:IsNotTuner(syncard) and --
			end
end
```
of course, ocgcore should also be updated to adjust this change.

nekrozar:
https://github.com/Fluorohydride/ygopro-scripts/commit/21e7c292e0843a349604b98fd43cfc2165d15030 updated.

PR #1098, opened 2018-10-06
temp fix

mercury233:
image

it can be very slow if the field is full of cyber dragons


purerosefallen:
~~I think you have to buy a new CPU~~

PR #1099, opened 2018-10-13
Add Auxiliary.GetMultiLinkedZone(tp)

KoishiMint:
Added Auxiliary.GetMultiLinkedZone(tp) for multi-linked zone(zone linked by two or more link monsters).
Special thanks to Bilibili user "我才不会写代码呢(I won't write code any more)".


purerosefallen:
You should open up another function in aux rather than use `local function` inside a function.
In addition, `a` `b` and `c` is not clear enough in code because no one else could read it.

PR #1100, opened 2018-10-13
Add Auxiliary.GetMultiLinkedZone(tp), FaceupLinkMonsterFilter(c)

KoishiMint:
Added Auxiliary.GetMultiLinkedZone(tp) for multi-linked zone(zone linked by two or more link monsters).
Special thanks to Bilibili user "我才不会写代码呢(I won't write code any more)".

### DIFF
--- a/utility.lua
+++ b/utility.lua
@@ -2121,14 +2121,13 @@ end
 --used for multi-linked zone(zone linked by two or more link monsters)
 function Auxiliary.GetMultiLinkedZone(tp)
 	local function f(c) return c:IsFaceup() and c:IsType(TYPE_LINK) end
-	--target linked zone
-	local tlz=0
 	--single-linked zone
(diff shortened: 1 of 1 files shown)


MonokuroInzanaito:
`local function f(c) return c:IsFaceup() and c:IsType(TYPE_LINK) end` What is this line for?
`mlz=slz & tlz | mlz` Which one goes first?

MonokuroInzanaito:
Shouldn't the 1st function be `Auxiliary.FaceupLinkMonsterFilter` or something like that?
`GetFaceupLinkMonster` is more likely a function returns a card group with all link monsters.

KoishiMint:
Updated function name and deleted the redundant line.
& goes first in order to calculate whether the monster being scanned increases multi-linked zone.

PR #1103, opened 2018-10-16
update Nekroz of Sophia

mercury233:
the new card _Impcantation Celebrated Birth_ can summon this card with materials level greater than 11


purerosefallen:
where is the script of Impcantation Celebrated Birth?
Also, if I were you, I would rewrite this procedure with SelectUnselect

PR #1117, opened 2018-11-20
fix & update Chrysalis, Contact and so on

nekrozar:
Fix this: Face-down Chrysalis send to grave by _Contact_.

https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=4&cid=6774
■効果処理時に、自分のモンスターゾーンに**表側表示で存在する**「C」と名のついたモンスターを全て墓地に送ります。

Update this: use `aux.IsCodeListed`

### DIFF
--- a/c16241441.lua
+++ b/c16241441.lua
@@ -12,6 +12,7 @@ function c16241441.initial_effect(c)
 	e1:SetOperation(c16241441.spop)
 	c:RegisterEffect(e1)
 end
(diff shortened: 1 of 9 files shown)


edo9300:
@nekrozar you should check the setcode of the summoned monster too, as on the official db in the effect explanation they said it must be a neo spacian monsters

PR #1129, opened 2018-12-27
use cache in subgroup check

purerosefallen:
Will store every combination during a single subgroup check process, which will make the whole check run faster.


mercury233:
why change bool `res` to number

PR #1130, opened 2019-01-03
A temp solution to 2 mat fusion

purerosefallen:


purerosefallen:
I still think we should do something on `Auxiliary.AddFusionProcMixRep`, because `Auxiliary.FCheckAdditional` should only check on the final material group, rather than every sub group of the material group.
However, @VanillaSalt wrote `Auxiliary.AddFusionProcMixRep` so complex that it was beyond my ability to read it. This is also why I didn't rewrite the procedure with `Group.SelectSubGroup`.

purerosefallen:
In addition, オスティナート(9113513) also needs update. ~But I'm too busy preparing my final exam~

PR #1134, opened 2019-01-15
add missing tp back in rituals

purerosefallen:


purerosefallen:
~Why not changing that Sophia~

PR #1187, opened 2019-06-02
add nil check to Comic Hand

szefo09:


DailyShana:
how can that varieble be nil?

szefo09:
that's a very good question I want to know myself. All I got from the logs is this: 
```
2|ygopro-s | {"name":"mycard","hostname":"raspberrypi","pid":4092,"level":30,"msg":"YGOPRO Debug: [string \"./script/c33453260.lua\"]:67: attempt to index a nil value (local 'tc')","time":"2019-06-02T08:23:02.827Z","v":0}
```

DailyShana:
since it's an optional effect, there is no reason to check nil, there must be some problem in other part

PR #1192, opened 2019-06-24
[WIP] scripts related to “refine the timing of adjust”

DailyShana:
Fluorohydride/ygopro-core#223


DailyShana:
the winning effect of CNo.88 should not be written in this way

PR #1193, opened 2019-06-25
fix: 大捕り物, 幻影霧剣

salix5:
@mercury233 @DailyShana @nekrozar @purerosefallen 
First of all, I apologize for all the inconvenience I have brought due to my carelessness.

56916a9fd4adc6c0fd61045f30e5b32c8a5d8e3d tried to fix Fluorohydride/ygopro#2200
I think this is because field effects are enabled before operation function executes in processor.cpp.
https://github.com/Fluorohydride/ygopro-core/blob/a1a33d155b23d8dcda1d44f416f4d748dd8451e3/processor.cpp#L4162

This is correct since 機殻の再星 has to check field effect first.
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=14155&keyword=&tag=-1

So the original Fog Blade do as follows:
1. EFFECT_DISABLE field effect is enabled before operation with null target.
2. The target of the field effect (which is null) are added into check list.
3. The target is created in operation  function.
4. After Fog Blade's effect resolves, the status of Candina is not refreshed since it is not in the check list. 

My opinion:
1. Continuous trap with a continuous target works like an equip card, so I think the disable check should treat them like equip cards.
This is what Fluorohydride/ygopro-core@7646ecfa532b690a6c966bc624a68e0a934eb05f tried to do.

2. Continuous trap with a continuous target and disable-related effects should put a EFFECT_FLAG_OWNER_RELATE effect on target cards instead of using field effects.
It will add the target into the check list correctly.
This is what 56916a9fd4adc6c0fd61045f30e5b32c8a5d8e3d  tried to do.

However, I made a big trouble because I did not notice that ordinary single effect does not check immunity.
https://github.com/Fluorohydride/ygopro-core/blob/a1a33d155b23d8dcda1d44f416f4d748dd8451e3/card.cpp#L2330
Only single effects with EFFECT_FLAG_SINGLE_RANGE will check immunity.

This commit fix this problem by adding EFFECT_FLAG_SINGLE_RANGE to Fog Blade.
Since Crackdown can catch face-down monsters and EFFECT_FLAG_SINGLE_RANGE works only on face-up monsters,  it check immunity in the condition functions.

About self-destroy conditions:
The conditions are modified in order to match most of the cards with similar effects.

Test replays:
https://www.sendspace.com/file/5ar638

arrest_xxxx
The interaction of Crackdown and disable, immune effects.
Unicorn starts from turn 4 due to some mistakes.

Candina_xxxx
The interaction of Candina and disable effects, 

full_armor_immune1
full_armor_immune2
The interaction of Blackwing Full Armor Master and Fog Blade, Crackdown.


DailyShana:
I want to fix this in Fluorohydride/ygopro-core#223
no matter a continuous effect target monster or not, it should applies at the beginning of chain solving

since the process of Fluorohydride/ygopro-core#223 may take long, 
for _Phantom Knights' Fog Blade_ it can be fixed by adding `Duel.AdjustInstantly(e:GetHandler())` after setting continuous target.

>Since Crackdown can catch face-down monsters and EFFECT_FLAG_SINGLE_RANGE works only on face-up monsters, it check immunity in the condition functions.

this can be done by adding a `EFFECT_FLAG_SET_AVAILABLE` and allowing `EFFECT_FLAG_SINGLE_RANGE` works for face-down monster when `EFFECT_FLAG_SET_AVAILABLE` exists.
This is more natural than the weird checking immunity in condition and `EFFECT_FLAG_SINGLE_RANGE` should not be removed.
because `EFFECT_FLAG_SINGLE_RANGE` is designed to simulate `EFFECT_TYPE_FIELD`, this is the reason why it checks immunity, not just a `EFFECT_FLAG_IGNORE_IMMUNE`.

DailyShana:
As continuous target is similar to equip target, it natural to have a new type `EFFECT_TYPE_TARGET` for effects affecting continuous target.
Changing around `EFFECT_FLAG_OWNER_RELATE` and `EFFECT_FLAG_SINGLE_RANGE` is endless

PR #1195, opened 2019-06-26
fix: 幻影霧剣, デモンズ・チェーン

salix5:
幻影霧剣, デモンズ・チェーン
1. Add Duel.AdjustInstantly() to avoid Fluorohydride/ygopro#2200.
2. The self-destroy conditions match similar effects.

大捕り物
~~Now "cannot attack or activate" applies to face-down monsters.~~
moved to another pull request

@DailyShana 
Add  Duel.AdjustInstantly() as workaround. Thank you for your advice.

Test replays:
https://www.sendspace.com/file/5ar638


DailyShana:
I think the taking control effect of _Crackdown_ should also be written as a continuous effect affecting effect target as the attack limit effect, and it is better to move all these effects to `initial_effect`.

salix5:
The modification of Crackdown will be moved to another pull request.
Please check if the workaround for Fog Blade is available.

PR #1199, opened 2019-07-08
update discard

nekrozar:
https://github.com/Fluorohydride/ygopro-core/pull/228


c14550855.lua
```diff
@@ -12,8 +12,14 @@ end
...
 end
+function c14550855.fselect(g,tp)
+	Duel.SetSelectedCard(g)
+	return Duel.CheckDiscardHand(tp,nil,0,REASON_DISCARD+REASON_EFFECT)
+end
 function c14550855.target(e,tp,eg,ep,ev,re,r,rp,chk)
-	if chk==0 then return Duel.GetFieldGroupCount(tp,LOCATION_HAND,0)>0
+	local g=Duel.GetDiscardHand(tp,REASON_DISCARD+REASON_EFFECT)
+	g:RemoveCard(e:GetHandler())
+	if chk==0 then return g:CheckSubGroup(c14550855.fselect,1,1,tp)
```
DailyShana: why not just `Duel.CheckDiscardHand(tp,nil,1,REASON_DISCARD+REASON_EFFECT,e:GetHandler())`

nekrozar: if `EFFECT_DISCARD_HAND_CHANGE`'s value is 0, `Duel.CheckDiscardHand` return true.
but, _Star Blast_'s FAQ is
> http://yugioh-wiki.net/index.php?%A1%D4%A5%B9%A5%BF%A1%BC%A1%A6%A5%D6%A5%E9%A5%B9%A5%C8%A1%D5#faq
> Ｑ：《魔力倹約術》でライフコストを支払わないことはできますか？
> Ａ：この場合、ライフポイントを払う事ができなくなり、このカードを発動する事自体できなくなります。(08/11/16)
so same card(e.g. _Linkslayer_) should use `Duel.SetSelectedCard`.

DailyShana: I see, the situation is similar to _Destiny HERO - Diamond Dude_ cannot use _Metamorphosis_

nekrozar: https://github.com/Fluorohydride/ygopro-scripts/pull/1199/commits/720b1e9601c817b18097e6d9c67ca0e48d048544 used `Duel.CheckDiscardHand`.

c14550855.lua
```diff
@@ -13,19 +13,30 @@ function c14550855.filter(c)
...
 	if ac==0 then return end
 	if ac>2 then ac=2 end
-	local ct=Duel.DiscardHand(tp,aux.TRUE,1,ac,REASON_DISCARD+REASON_EFFECT)
-	Duel.BreakEffect()
+	local min,max=Duel.GetDiscardHandChangeCount(tp,REASON_EFFECT,1,ac)
+	if max<=0 then return end
+	if min<=0 then min=1 end
+	if Duel.DiscardHand(tp,nil,min,max,REASON_DISCARD+REASON_EFFECT)==0 then return end
+	local ct=Duel.GetOperatedGroup():GetCount()
+	if ct<=0 then return end
```
DailyShana: also, why not use return value of `Duel.DiscardHand` directly

nekrozar: if `Duel.DiscardHand` called `EFFECT_DISCARD_HAND_REPLACE`, `Duel.DiscardHand` return 1.
so, `Duel.GetOperatedGroup` is used (similar to `Card.RemoveOverlayCard`).

DailyShana: if `EFFECT_SENDTO_REPLACE` makes no cards moved, `Duel.SendtoHand`, etc return 0.
i.e., the return value should mean the count of actually moved cards.

nekrozar: but, `EFFECT_OVERLAY_REMOVE_REPLACE` must be returned 1.
https://github.com/Fluorohydride/ygopro/issues/2220

DailyShana: Effects like _Xyz Override_ is replacing one of overlay cards to be removed, i.e. the replacing is treated as 1 removed overlay cards, so the replacing is counted in then return value of `RemoveOverlayCard`.
But effects of ウィッチクラフト・バイストリート is replacing the whole operation of discard hand, i.e., no discard happened, it is replaced, so it should return 0 in `Duel.DiscardHand`.

DailyShana: so in this case, `EFFECT_DISCARD_HAND_REPLACE` should be treated similarly as `EFFECT_SENDTO_REPLACE` but not `EFFECT_OVERLAY_REMOVE_REPLACE`

nekrozar: https://github.com/Fluorohydride/ygopro-scripts/pull/1199/commits/d17594bad4f52de85f62513ef1fda1e87378fa03 updated.

PR #1205, opened 2019-07-16
Possible solution for Tune Magician

purerosefallen:


DailyShana:
this line https://github.com/Fluorohydride/ygopro-scripts/pull/1205/files#diff-e2dcb2b3468cae90b8ab3113481f4879R1060 also need to exclude contact fusion

PR #1245, opened 2019-11-16
add aux.RCheckAdditional, aux.RGCheckAdditional

nekrozar:


purerosefallen:
Seems good

nekrozar:
_Megalith_ use `aux.RCheckAdditional`.
So, `gc` is unnecessary.

PR #1251, opened 2019-11-27
update selecting cards with different name/level etc.

mercury233:


c10424147.lua
```diff
@@ -80,7 +80,7 @@ function c10424147.sptg2(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
 	if chkc then return false end
 	local g=Duel.GetMatchingGroup(c10424147.spfilter3,tp,LOCATION_MZONE+LOCATION_GRAVE,0,nil,e)
 	if chk==0 then return Duel.GetLocationCountFromEx(tp)>0
-		and g:CheckSubGroup(aux.dncheck,3,3)
+		and g:GetClassCount(Card.GetCode)>=3
```
nekrozar: why?

mercury233: may have better performance

nekrozar: however, _Simorgh Onslaught_ hasn't changed(maybe other scripts).

mercury233: _Simorgh Onslaught_ is updated.

PR #1257, opened 2019-12-08
fix confirm

nekrozar:
faq
faq

### DIFF
--- a/c10698416.lua
+++ b/c10698416.lua
@@ -65,5 +65,6 @@ function c10698416.spop(e,tp,eg,ep,ev,re,r,rp)
 	if sg then
 		Duel.SpecialSummon(sg,0,tp,tp,false,false,POS_FACEDOWN_DEFENSE)
 		Duel.ConfirmCards(1-tp,sg)
+		Duel.ShuffleSetCard(sg)
 	end
 end
--- a/c15710054.lua
+++ b/c15710054.lua
@@ -62,5 +62,6 @@ function c15710054.spop(e,tp,eg,ep,ev,re,r,rp)
 	if sg then
 		Duel.SpecialSummon(sg,0,tp,tp,false,false,POS_FACEDOWN_DEFENSE)
 		Duel.ConfirmCards(1-tp,sg)
+		Duel.ShuffleSetCard(sg)
 	end
 end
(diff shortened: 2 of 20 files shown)


mercury233:
the player should be able to select the zone (linked or not) for each monster, so shuffle is not a good idea

nekrozar:
however, `Duel.SSet` use `MSG_SHUFFLE_SET_CARD`.
so, these problem happen `Duel.SSet`.

mercury233:
no, I let `Duel.SSet` rearrange the cards with selected order, not shuffle

nekrozar:
so should we fix ocgcore?

mercury233:
it is hard to change the special summon procedure...

edo9300:
Wouldn't it be more practical to just make the opponent confirm the cards before setting?

mercury233:
@edo9300 It may work for summoning from deck or hand, but will not work for summoning from grave, because the cards in grave are public, so the movement is public too

edo9300:
Hmm, you're right, an approach then would be to add an extra parameter in the procedures to "fake" the shuffling for the opponent, even if idk how that would be doable with spsummon step

DailyShana:
Let's merge the fix of _Subterror Behemoth Fiendess_ and _Galaxy Satellite Dragon_ first.

nekrozar:
https://github.com/Fluorohydride/ygopro-scripts/commit/c7532f0f700db2fd826ae09e5abc7a86410eb707 merged _Subterror Behemoth Fiendess_ and _Galaxy Satellite Dragon_.

nekrozar:
https://github.com/Fluorohydride/ygopro-core/blob/master/libduel.cpp#L446
https://github.com/Fluorohydride/ygopro-scripts/pull/1257/commits/50eeadb016e60d29b1468760d0b99210b9d232a9#diff-05562c57b68375d1e918c487658a29f7R70
https://github.com/Fluorohydride/ygopro-scripts/pull/1257/commits/50eeadb016e60d29b1468760d0b99210b9d232a9#diff-5da6591e3b67811243447d4bf240a8b3R53
And, if group's cards has only 1 card in `Duel.SSet`, setting cards don't confirm.
So, _Abyss Script - Romantic Terror_ and _Ultimate Fire Formation - Seito_ should be check counts.
It would be better to confirm only 1 card by default.

mercury233:
```lua
Duel.ConfirmCards(1-tp,sg)
Duel.ShuffleDeck(tp)
Duel.SpecialSummon(sg,0,tp,tp,false,false,POS_FACEDOWN_DEFENSE)
```
@edo9300 this will not actually work because `MSG_MOVE` contain the code of the cards

edo9300:
Shouldn't the code be 0ed by the server when sending the message with facedown cards?

mercury233:
@nekrozar https://github.com/Fluorohydride/ygopro-core/pull/248 updated
@edo9300 actually it didn't

PR #1258, opened 2019-12-08
use event player

mallu11:


DailyShana:
Can you explain the detail issue briefly?

mallu11:
> Can you explain the detail problem briefly?

Before these changes,if your opponent control your I:P and use it to link summon a monster,that monster can be destroyed by your effect.In fact,the "opponent" in the effect "can not be destroyed by your opponent's effect" is the summon player's opponent.
A same example is <i>Windwitch - Snow Bell</i> which uses event player.

DailyShana:
ジャンク・ジャイアント and 聖騎士エクター・ド・マリス seem no such problem according test

mallu11:
> ジャンク・ジャイアント and 聖騎士エクター・ド・マリス seem no such problem according to test

I changes them without testing.If you think the change is unnecessary,mayde revert the code of them.

PR #1267, opened 2019-12-12
use aux.GCheckAdditional

mercury233:
better performance
[credit to](https://github.com/Fluorohydride/ygopro-scripts/commit/99a7162f1deffcfb02ce2a7f03ee6e76ba2cbba7#commitcomment-36390714) @nekrozar


nekrozar:
it is better to change `aux.dlvcheck` and so on.

PR #1269, opened 2019-12-12
add Group.SelectSubGroupEach etc.

mercury233:


nekrozar:
It would be better to add `min` and `max` by _Utopic Onomatopoeia_.

mercury233:
@nekrozar what about this way

nekrozar:
should we write these long code?

PR #1271, opened 2019-12-14
update add table

nekrozar:


DailyShana:
want to know what problem is encountered

nekrozar:
it is better to add different value only.

DailyShana:
related: #860

DailyShana:
to use table as a set, the practical way is 
```lua
set={}
set[element]=true
```

DailyShana:
the name `TableInsertValue` is misleading since it do the different things to `table.insert`

I suggest storing element in key, it's simple and direct and in fact `c.material_count` was just used as loop count in for statement so can be safely removed

nekrozar:
But `TableInsertValue` is used `c.card_code_list`.

DailyShana:
card_code_list can also be change in this way

nekrozar:
hmm, could you tell me `c.material_count`'s using code?

DailyShana:
https://github.com/Fluorohydride/ygopro-scripts/commit/68bc29f714effac7dd37e21fa57ed81647771d94

nekrozar:
ok understood.
https://github.com/Fluorohydride/ygopro-scripts/pull/1271/commits/10b372cd843e4f713a10fe814fef24abf04ada20 updated.

utility.lua
```diff
@@ -1015,20 +1014,19 @@ function Auxiliary.AddFusionProcMix(c,sub,insf,...)
 					return false
 			end
 			for _,fcode in ipairs(val[i]) do
-				if type(fcode)~='function' then table.insert(mat,fcode) end
+				if type(fcode)~='function' and not mat[fcode] then mat[fcode]=true end
 			end
 		else
 			fun[i]=function(c,fc,sub) return c:IsFusionCode(val[i]) or (sub and c:CheckFusionSubstitute(fc)) end
-			table.insert(mat,val[i])
+			if not mat[val[i]] then mat[val[i]]=true end
```
DailyShana: I think `if not ...` is unnecessary

nekrozar: https://github.com/Fluorohydride/ygopro-scripts/pull/1271/commits/62e709b70a275404e5e11c48c298569c8756cd18 updated.

PR #1273, opened 2019-12-16
fix equip

mercury233:
fix this: if the target card is immune to the equip effect and the monster is face-down when the effect resolves, the target card shouldn't be `Duel.SendtoGrave(tc,REASON_RULE)`


DailyShana:
I think it should be done in ocgcore

PR #1275, opened 2019-12-17
clear operation info when copying effect

DailyShana:
fix https://github.com/Fluorohydride/ygopro/issues/2253


c22628574.lua
```diff
@@ -40,7 +40,7 @@ function c22628574.target(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
 	e:SetCategory(te:GetCategory())
```
mallu11: The "e:SetCategory(te:GetCategory())" is unnecessary which may make the effect can be chained by ash blossom.(There are more cards to change)

PR #1277, opened 2019-12-17
Fix "Power Wall"

Lyris12:
Interaction between "Power Wall" & "Blue-Eyes Chaos MAX Dragon"

### DIFF
--- a/c76403456.lua
+++ b/c76403456.lua
@@ -20,10 +20,6 @@ function c76403456.target(e,tp,eg,ep,ev,re,r,rp,chk)
 	Duel.SetOperationInfo(0,CATEGORY_DECKDES,nil,0,tp,g)
 end
 function c76403456.activate(e,tp,eg,ep,ev,re,r,rp)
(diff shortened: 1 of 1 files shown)


DailyShana:
we can't change in this way, if the cards sent to grave have "if this card is sent to graveyard" effect, such effect will be activated in the wrong timing

Lyris12:
Well, currently, **8** cards (instead of 16) are discarded from the Deck with "Power Wall" to reduce **8000** piercing battle damage from "Blue-Eyes Chaos MAX Dragon" to 0. Is there another way to correct this interaction?

DailyShana:
I know what you are fixing but currently it's hard to resolve it

edo9300:
@DailyShana they way we fixed that on percy is by using external functions to set/retrieve the battle damage instead of erlying on the argument pased in the effect, the issue here is that all th effects are raised at teh same time, all with teh same value, so that remains the same when resolving all the related effects

DailyShana:
@edo9300 thank you for you advise. I plan to update EFFECT_PIERCE to support double pierce damage

PR #1285, opened 2019-12-23
fix double piercing damage

DailyShana:
https://github.com/Fluorohydride/ygopro-core/pull/259

fix Fluorohydride/ygopro-core#97 close Fluorohydride/ygopro-scripts#1277


ClementLouis:
Tell me if i missed something (if so I'm sorry) but from what i understood about the change, I think you missed an interaction
the following ruling explain that Damage from Blue-eyes Chaos MAX (that have "double piercing damage") can also be doubled by another card effect (Action Magic Full turn is used in the exemple), so for me, it look like there is Double_Piercing that should be different from DOUBLE_DAMAGE since you can't stack double damage on top of another double damage, but you can stack an double damage with an double piercing (idk if I'm clear enough XP)
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=12895&keyword=&tag=-1&request_locale=ja


We can suppose it should work similarly with halving damage, if an "half piercing" existe one day, then it would stack with an "half damage" effect

DailyShana:
`DOUBLE_DAMAGE` is not a new effect type, when `DOUBLE_DAMAGE` is set as **value** of effect of type `EFFECT_PIERCE`, it means double piecing battle damage.

PR #1287, opened 2019-12-24
update fusion spell check

mercury233:
fix https://github.com/Fluorohydride/ygopro-scripts/pull/1244#issuecomment-554725326
require https://github.com/Fluorohydride/ygopro-scripts/pull/1286

### DIFF
--- a/c21140872.lua
+++ b/c21140872.lua
@@ -48,6 +48,9 @@ function c21140872.initial_effect(c)
 	c:RegisterEffect(e3)
 end
 c21140872.material_setcode=0x3b
+function c21140872.red_eyes_fusion_check(tp,sg,fc)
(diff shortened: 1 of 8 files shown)


edo9300:
This is wrong, with the available rulings, you have to use (in case of red eyes fusion) the "red-eyes" monster as the listed "red-eyes" monster, it's not enough having a red-eyes in the mats

mercury233:
@edo9300 take Slash Dragon for example, it do check the fusion material contain a red-eye B dragon and a warrior, and the 2 cards must be different

edo9300:
Hmm, I'm not a fan of this approach, as it's very specific, _but it seems to work_ anyways, it's not only red eyes fusion that has to be updated, miracle synchro fusion should as well, and iirc there are other similar cards that fall under that category.

mercury233:
I found that we only need to update those fusion monsters: The designated part of material can be replaced, and the rest part can also match the requirement of the fusion spell.

PR #1298, opened 2020-01-04
update EVENT_BE_MATERIAL + REASON_XYZ

mercury233:
after https://github.com/Fluorohydride/ygopro-core/pull/258 , `tp` and `SetCountLimit` will work correctly


DailyShana:
I will fix this

PR #1299, opened 2020-01-05
Fix "Ghost Sister & Spooky Dogwood" & "Blazing Mirror Force"

Lyris12:
Cards which depend on amount of LP lost (by damage) or gained got glitched up because `Duel.Damage` and `Duel.Recover` are returning nil.
- "Ghost Sister" halved her user's LP at the end of the turn, even if they gained LP with her effect that turn
- "Blazing Mirror Force" did not damage the opponent after damaging its user

### DIFF
--- a/c60643553.lua
+++ b/c60643553.lua
@@ -68,7 +68,9 @@ end
 function c60643553.lpop1(e,tp,eg,ep,ev,re,r,rp)
 	local lg=eg:Filter(c60643553.cfilter,nil,1-tp)
 	local rnum=lg:GetSum(Card.GetAttack)
-	if Duel.Recover(tp,rnum,REASON_EFFECT)<1 then return end
+	local lp=Duel.GetLP(tp)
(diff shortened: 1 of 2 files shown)


DailyShana:
>because `Duel.Damage` and `Duel.Recover` are returning nil

but actually no such problem

Lyris12:
...I'll put this elsewhere, then.

Lyris12:
So, I'm being directed back here instead of said elsewhere. Does anything happen at the end of the turn which "Ghost Sister"'s effect gave her user LP? Does the opponent get damaged by "Blazing Mirror Force"?

DailyShana:
These two functions has no problem, basically. 
So you should describe the situation in detail, and even though these two cards can not work properly in some situation, we should go to fix Duel.Damage and Duel.Recover but not in this way.

DailyShana:
can't reproduce

PR #1311, opened 2020-01-20
fix for disable trapmonster

mallu11:


mallu11:
And maybe can change to:
`function Auxiliary.disfilter1(c,trap)`
`return c:IsFaceup() and (not c:IsDisabled() or (not trap or c:IsType(TYPE_TRAPMONSTER))) and (not c:IsType(TYPE_NORMAL) or c:GetOriginalType()&TYPE_EFFECT~=0)`
`end`
The `trap` is a bool parameter.

mallu11:
And maybe can change to:
`function Auxiliary.disfilter1(c,trap)`
`return c:IsFaceup() and not (c:IsDisabled() and (not trap or not c:IsType(TYPE_TRAPMONSTER))) and (not c:IsType(TYPE_NORMAL) or c:GetOriginalType()&TYPE_EFFECT~=0)`
`end`
The `trap` is a bool parameter.

mercury233:
~~`disfilter2` is better than `true`~~

PR #1312, opened 2020-01-20
fix 神聖魔皇后セレーネ

mallu11:


DailyShana:
equip card that was a monster automatically has TYPE_SPELL

mallu11:
> equip card that was a monster automatically has TYPE_SPELL

But before the change,the filter cann't get the face-down equip card.

PR #1317, opened 2020-01-26
Fix "Cloudian - Poison Cloud"

Lyris12:
`if it was face-up at the start of the Damage Step` is absent from the OCG version when compared to the TCG version, but the script reflects the latter version: https://yugipedia.com/wiki/Cloudian_-_Poison_Cloud#Other_languages

I fixed the script of "Cloudian - Poison Cloud" to work like in the OCG.

### DIFF
--- a/c83982270.lua
+++ b/c83982270.lua
@@ -13,7 +13,7 @@ function c83982270.initial_effect(c)
 end
 function c83982270.condition(e,tp,eg,ep,ev,re,r,rp)
 	local c=e:GetHandler()
-	return c:IsLocation(LOCATION_GRAVE) and c:IsReason(REASON_BATTLE) and c:IsPreviousPosition(POS_FACEUP)
+	return c:IsLocation(LOCATION_GRAVE) and c:IsReason(REASON_BATTLE)
 end
 function c83982270.target(e,tp,eg,ep,ev,re,r,rp,chk)
 	if chk==0 then return true end


DailyShana:
the script just say the following:
フィールド上に**表側表示**で存在するこのカードが戦闘によって破壊され墓地へ送られた時
When this card that was face-up on field is destroyed by battle and sent to the Graveyard

nothing about start of damage step is written in the script

Lyris12:
Oh. I guess Yugipedia had mistranslated the card. Thanks for the clarification, though

Dinoguy1000:
A [current ruling](https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=4&cid=6198&request_locale=ja) reads:
> ■「雲魔物－ポイズン・クラウド」自身が戦闘で破壊され墓地へ送られたダメージステップ終了時に必ず発動する効果です。
> *This effect always activates at the end of a Damage Step in which "Cloudian - Poison Cloud" itself was destroyed by battle and sent to the Graveyard.*

So in the OCG the effect will activate even if it was face-down at the start of the Damage Step.

The text in question literally translates to `When this face-up card is destroyed by battle and sent to the Graveyard`.

In the past, this had been ruled as meaning that the card had to be face-up at the start of the Damage Step in which it was destroyed by battle. However, now this text is ruled as simply meaning that the card has to be face-up when it is destroyed, which—in the case that it specifically has to be destroyed by battle—renders the "face-up" qualifier redundant.

The old ruling caused `and it was face-up at the start of the Damage Step` clauses to be added to the PSCT of several TCG cards ("Cyber Phoenix", "Fox Fire", "Evilswarm Zahak", "Cloudian - Poison Cloud"). Those that have been reprinted in the TCG since the ruling change ("Cyber Phoenix", "Evilswarm Zahak") removed that clause in their reprinted text. Modern prints of cards that have effects that activate when destroyed by battle while face-up just remove the "face-up" qualifier entirely ("Cyber Phoenix", "Jade Knight").

The one exception is "House Duston", whose Japanese text was [specifically updated](https://yugipedia.com/wiki/Card_Errata:House_Duston) around the time of the ruling change in order to specifically state that it does have to be face-up at the start of the Damage Step. Prior to its AT17 reprint, it had the same text as the others.

PR #1328, opened 2020-02-07
fix 天威龍－サハスラーラ

mallu11:


mallu11:
According to the answer of KONAMI,the special summon is unrelating the target monster.

PR #1329, opened 2020-02-09
fix 閃刀姫

mallu11:


mallu11:
These effect only can be activated if the card is already in graveyard,so if the handler card was send to graveyard in the same event,the effect can't be activated.

PR #1330, opened 2020-02-09
fix 百景戦都ゴルディロックス

mallu11:


mallu11:
fix the missing categoryinfo that makes it can't be chained by some effects.

c24521754.lua
```diff
@@ -38,17 +40,32 @@ end
...
 	Duel.Hint(HINT_SELECTMSG,tp,HINTMSG_TOZONE)
 	local fd=Duel.SelectDisableField(tp,1,LOCATION_MZONE,0,0)
 	Duel.Hint(HINT_ZONE,tp,fd)
-	e:SetLabel(math.log(fd,2))
+	local seq=math.log(fd,2)
+	e:SetLabel(seq)
+	local pseq=c:GetSequence()
+	if pseq>seq then pseq,seq=seq,pseq end
+	local dg=Group.CreateGroup()
+	dg:KeepAlive()
```
DailyShana: `KeepAlive` is unnecessary

c24521754.lua
```diff
@@ -38,17 +40,32 @@ end
...
 end
-function c24521754.seqfilter(c)
-	local tp=c:GetControler()
-	return c:IsFaceup() and c:GetSequence()<5 and Duel.GetLocationCount(tp,LOCATION_MZONE,tp,LOCATION_REASON_CONTROL)>0
+function c24521754.seqcon(e,tp,eg,ep,ev,re,r,rp)
+	return e:GetHandler():GetSequence()<5
 end
 function c24521754.seqtg(e,tp,eg,ep,ev,re,r,rp,chk)
 	local c=e:GetHandler()
 	if chk==0 then return Duel.GetLocationCount(tp,LOCATION_MZONE)>0 end
```
DailyShana: should be
```suggestion
	if chk==0 then return Duel.GetLocationCount(tp,LOCATION_MZONE,PLAYER_NONE,0)>0 end
```

PR #1332, opened 2020-02-11
fix フューチャー・ドライブ

mallu11:


mallu11:
fix:the effect is treated as the monster's effect.In fact,even if the control of the monster is changed,the effect's owner is always the player who activates this card.

PR #1337, opened 2020-02-14
Added LOCATION_PZONE

TheBoos2569:
The location constant was added, because it seems LOCATION_ONFIELD does not include PZONE for whatever reason. For example, Lunalight Tiger will not trigger when destroyed in the PZONE. This is a temporary fix to the 3 cards: Lunalight Tiger, Performage Plushfire and Raremetalfoes Bismugear, until the constant is changed.


mercury233:
this problem doesn't exist in my test, please make sure you are using the ocgcore from here

PR #1339, opened 2020-02-18
add Auxiliary.bpcon and Auxiliary.dscon

mercury233:
There are over 200 scripts repeating those codes


DailyShana:
please use intuitive names

mercury233:
it is annoying to write long function names

DailyShana:
functions in `aux` are just like standard library in c++, they are part of the script api

mercury233:
we had many short function names and variable names already, so for keeping API style we should use more in future

utility.lua
```diff
@@ -2090,6 +2090,14 @@ end
 function Auxiliary.exccon(e)
 	return Duel.GetTurnCount()~=e:GetHandler():GetTurnID() or e:GetHandler():IsReason(REASON_RETURN)
 end
+--condition of checking battle phase availability
+function Auxiliary.bpcon(e,tp,eg,ep,ev,re,r,rp)
+	return Duel.IsAbleToEnterBP() or (Duel.GetCurrentPhase()>=PHASE_BATTLE_START and Duel.GetCurrentPhase()<=PHASE_BATTLE)
```
DailyShana: if `Duel.GetCurrentPhase()>=PHASE_BATTLE_START and Duel.GetCurrentPhase()<=PHASE_BATTLE` is always needed, we should consider merge is condition into `Duel.IsAbleToEnterBP`

PR #1347, opened 2020-02-23
fix location count check of zone moving effects

mercury233:
https://github.com/Fluorohydride/ygopro-scripts/commit/b1fae4b294f1a8b3e04846c7e22e0369af866542


DailyShana:
thank you

PR #1350, opened 2020-03-03
fix クロスローズ・ドラゴン

mallu11:


mallu11:
the effect can't be activated when only this matching card in graveyard.

PR #1351, opened 2020-03-03
fix 飢鰐竜アーケティス

mallu11:


mallu11:
use the `EFFECT_MATERIAL_CHECK` to check the material when setmaterial but not check the material after special summon success.

PR #1353, opened 2020-03-05
fix 繁華の花笑み

mallu11:


mallu11:
fix the effect can't be chained by Ash Blossom.

PR #1357, opened 2020-03-07
fix

mallu11:


mallu11:
the effect maybe can affect the handler card, but it can't reset when the card is disabled.

PR #1383, opened 2020-04-08
fix 魂のしもべ

mallu11:
shouldn't shuffle deck if the card you select is not in deck.


mercury233:
after you looked at your deck, you should always shuffle it, and the correct time is before the time that the selected card is placed on the top of the deck.

mallu11:
like other effects, it only shows the matching cards in your deck and doesn't show all the card. It had better hide the sequence or rearrange the showing sequence when you select the card in deck.

mercury233:
anyway, your change will make the deck be shuffled after you place a card from grave

mallu11:
It uses `Duel.SendtoDeck(tc,nil,0,REASON_EFFECT)`, and I didn't change it. The deck won't shuffle if you select the card from grave.

mercury233:
`Duel.SelectMatchingCard` will

PR #1400, opened 2020-05-16
Fix "Breaking of the World"

Lyris12:
First option incorrectly said the description of the Level Change effect


purerosefallen:
I'm planning to migrate those multi-select effects into one

mercury233:
I prefer to change `e2:SetDescription(aux.Stringid(69217334,0))` so the databases don't need to be changed

Lyris12:
That's fine with me, too.

PR #1407, opened 2020-05-21
Reset registering effect of Salamangreat Burning Shell if summon is negated

edo9300:
If Salamangreat Burning Shell is activated as chain link 1, the link summon can then be negated by cards like solemn judgment, in which case EVENT_SPSUMMON_SUCCESS would never be called, and thus never resetted, this making it be applied the next time that card would be special summoned.


mercury233:
Maybe something like メタファイズ・ファクター will work?

edo9300:
This is different, factor adds an extra summon procedure to the monster, while this issue is related to the change done to the summon functions behaviour during an effect, so they're completely different scenarios.

PR #1408, opened 2020-05-22
fix Salamangreat Burning Shell

mercury233:
fix https://github.com/Fluorohydride/ygopro-scripts/pull/1407


mercury233:
@edo9300

c28534130.lua
```diff
@@ -58,6 +58,7 @@ function c28534130.activate(e,tp,eg,ep,ev,re,r,rp)
 		local e3=Effect.CreateEffect(c)
 		e3:SetType(EFFECT_TYPE_SINGLE+EFFECT_TYPE_CONTINUOUS)
 		e3:SetCode(EVENT_SPSUMMON_SUCCESS)
+		e3:SetReset(RESET_EVENT+RESETS_STANDARD-RESET_TOFIELD)
```
DailyShana, line 61: and reset overlay, reset mschange

edo9300:
Yeah, this is cleaner, even tho, i think it would be nice if there was teh possibility to set a callback function that would be called after the summon is performed in the core, as that'd make those scenarios easy to handle, do you think that's a feasible thing @DailyShana?

PR #1414, opened 2020-05-27
fix equip spell card about changing control

mallu11:
according to KONAMI's answer, the equip spell card can't target the monster that can't change control.


mercury233:
https://github.com/Fluorohydride/ygopro-scripts/pull/1227

mallu11:
i'm sorry about the wrong change, but we should fix it now.

PR #1416, opened 2020-05-31
Fix "Barricadeborg Blocker"

Lyris12:
Currently: This card can't activate the effect to add a Continuous or Field Spell from GY to hand if it's not there.
Fix: It should be allowed to activate even if it's not in the GY. If at the EP, there's no target, then nothing is added.

Source: https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=19782&request_locale=ja


mercury233:
still confirming

mercury233:
can't be confirmed, but can activate is better than can't in a sim.

PR #1417, opened 2020-06-02
Fix "Legendary Six Samurai - Mizuho"

Lyris12:
Was able to Tribute face-down "Six Samurai" monsters


mercury233:
The ocg text didn't say "face-up"

PR #1418, opened 2020-06-03
Fix "Mausoleum of the Emperor"

Lyris12:
When using the effect of the field spell "Mausoleum of the Emperor" to Summon a monster, the opponent was able to activate a "Forced Back" or "Solemn Judgment" to negate the summon when they should not be able to, due to ruling mechanics.


mallu11:
why do you think "Solemn Judgment" should not be able to negate it? in fact, when using the effect of "Ultimate Offering" to summon a monster(as chain 1), the "Solemn Judgment" can negate the summon.

edo9300:
The existing rulings state that you CAN chain a card to negate that summon ```
Ｑ：この効果を適用した通常召喚はチェーンブロックを作りますか？
Ａ：はい、チェーンブロックを作ります。
　　なお、この召喚はチェーンブロックを作りますが《昇天の角笛》等の無効化対象になります。(10/03/04)
Q: Do normal summons that apply this effect make chain blocks?
A: Yes, I will make a chain block.
Note that this summon will make a chain block, but it will be subject to invalidation such as 《Ascension Horn》. (10/03/04)
Ｑ：このカードの効果で通常召喚を行うタイミングはこのカードの効果処理中ですか、それとも効果処理終了直後ですか？
Ａ：効果処理終了後にすぐに通常召喚を行います。(12/09/12)
Q: Is the timing of normal summoning with the effects of this card in effect processing of this card, or right after the effects processing ends?
A: Normal summon will be performed immediately after the effect processing is completed. (12/09/12)
Source: https://yugioh-wiki.net/index.php?%A1%D4%BB%E0%B9%C4%C4%EB%A4%CE%CE%CD%CA%E8%A1%D5
```
Source (https://yugioh-wiki.net/index.php?%A1%D4%BB%E0%B9%C4%C4%EB%A4%CE%CE%CD%CA%E8%A1%D5)

If then you have a more recent ruling stating that you CANNOT negate that summon, then provide it

Lyris12:
> why do you think "Solemn Judgment" should not be able to negate it? in fact, when using the effect of "Ultimate Offering" to summon a monster(as chain 1), the "Solemn Judgment" can negate the summon.

Well, the PSCT of "Ultimate Offering" says the following:
`During your Main Phase or your opponent's Battle Phase: You can pay 500 Life Points; immediately after this effect resolves, Normal Summon/Set 1 monster.`
Note the part, `immediately after this effect resolves`.

edo9300:
@Lyris12 regardless of the text having psct or not, there's a ruling stating that you can negate it, and rulings take priority over texts for obvious reasons.

Lyris12:
Alright, then...

PR #1433, opened 2020-06-23
fix 捕食植物ヴェルテ・アナコンダ

mercury233:
https://github.com/Fluorohydride/ygopro-scripts/pull/1431 need this card to be fixed too


DailyShana:
as effect copying other effect alway use its own `e,tp` in `cost`, `target`, `operation`, `Card.CheckActivateEffect` should be updated.

PR #1438, opened 2020-06-29
temp fix Sin パラドクス・ドラゴン

salix5:
@DailyShana @mercury233 
Problem
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=10978&keyword=&tag=-1
[bug_paradox.zip](https://github.com/Fluorohydride/ygopro-scripts/files/4847177/bug_paradox.zip)
Now: 2nd effect of Sin パラドクス・ドラゴン can be activated if Sin World is not on the field.
Correct: 2nd effect of Sin パラドクス・ドラゴン cannot be activated if Sin World is not on the field.

### DIFF
--- a/c8310162.lua
+++ b/c8310162.lua
@@ -28,7 +28,7 @@ function c8310162.descon(e)
 	return not Duel.IsEnvironment(27564031)
 end
 function c8310162.spcon(e,tp,eg,ep,ev,re,r,rp)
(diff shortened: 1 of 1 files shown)


DailyShana:
this should be handled by the new rule update in core.
this problem is probably caused by raising `EVENT_SPSUMMON_SUCCESS` when this monster is in grave.
So, it's a problem of the self destroy and should be fix in core

PR #1444, opened 2020-07-06
fix 魔導書庫クレッセン

mallu11:
the selection is random.


mercury233:
`RandomSelect` need rework

PR #1446, opened 2020-07-09
fix special summon that is treated as fusion summon

mallu11:
`c:SetMaterial(nil)` doesn't work.

### DIFF
--- a/c12289247.lua
+++ b/c12289247.lua
@@ -143,8 +143,12 @@ function c12289247.hnop(e,tp,eg,ep,ev,re,r,rp)
 	if not aux.MustMaterialCheck(nil,tp,EFFECT_MUST_BE_FMATERIAL) then return end
 	Duel.Hint(HINT_SELECTMSG,tp,HINTMSG_SPSUMMON)
 	local g=Duel.SelectMatchingCard(tp,c12289247.hnfilter,tp,LOCATION_EXTRA,0,1,1,nil,e,tp,nil)
-	if g:GetCount()>0 then
-		Duel.SpecialSummon(g,SUMMON_TYPE_FUSION,tp,tp,false,false,POS_FACEUP)
-		g:GetFirst():CompleteProcedure()
+	local tc=g:GetFirst()
+	if tc then
+		local mg=Group.CreateGroup()
+		tc:SetMaterial(mg)
+		if Duel.SpecialSummon(tc,SUMMON_TYPE_FUSION,tp,tp,false,false,POS_FACEUP) then
+			tc:CompleteProcedure()
+		end
 	end
 end
--- a/c12289247.lua
+++ b/c12289247.lua
@@ -147,7 +147,7 @@ function c12289247.hnop(e,tp,eg,ep,ev,re,r,rp)
 	if tc then
 		local mg=Group.CreateGroup()
 		tc:SetMaterial(mg)
-		if Duel.SpecialSummon(tc,SUMMON_TYPE_FUSION,tp,tp,false,false,POS_FACEUP) then
+		if Duel.SpecialSummon(tc,SUMMON_TYPE_FUSION,tp,tp,false,false,POS_FACEUP)~=0 then
 			tc:CompleteProcedure()
 		end
 	end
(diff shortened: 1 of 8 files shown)


mercury233:
don't work for which card?

mallu11:
> don't work for which card?

the `Lyrilusc - Independent Nightingale`.
(1)use `Polymerization` to fusion summon it.
(2)use `Compulsory Evacuation Device` to return it to extra deck.
(3)use `Instant Fusion` to summon it, it can activate its first effect.

mallu11:
there are other cards should be changed.

mercury233:
any replay? isn't it a bug of the bird?

edo9300:
Are you sure about set material not working with nil?
```
int32 scriptlib::card_set_material(lua_State *L) {
	check_param_count(L, 2);
	check_param(L, PARAM_TYPE_CARD, 1);
	card* pcard = *(card**) lua_touserdata(L, 1);
	if(!lua_isnil(L, 2)) {
		check_param(L, PARAM_TYPE_GROUP, 2);
		group* pgroup = *(group**) lua_touserdata(L, 2);
		pcard->set_material(&pgroup->container);
	} else
		pcard->set_material(0);
	return 0;
}```
It should work properly with both the parameter being nil and with no parameters at all

mallu11:
> Are you sure about set material not working with nil?
> 
> ```
> int32 scriptlib::card_set_material(lua_State *L) {
> 	check_param_count(L, 2);
> 	check_param(L, PARAM_TYPE_CARD, 1);
> 	card* pcard = *(card**) lua_touserdata(L, 1);
> 	if(!lua_isnil(L, 2)) {
> 		check_param(L, PARAM_TYPE_GROUP, 2);
> 		group* pgroup = *(group**) lua_touserdata(L, 2);
> 		pcard->set_material(&pgroup->container);
> 	} else
> 		pcard->set_material(0);
> 	return 0;
> }
> ```
> 
> It should work properly with both the parameter being nil and with no parameters at all

in fact, the setmaterial(nil) works properly, but the `EFFECT_MATERIAL_CHECK` works incorrect when the parameter is nil. So we can change it to setmaterial(Group.CreateGroup()) or fix the `EFFECT_MATERIAL_CHECK`.

edo9300:
> > Are you sure about set material not working with nil?
> > ```
> > int32 scriptlib::card_set_material(lua_State *L) {
> > 	check_param_count(L, 2);
> > 	check_param(L, PARAM_TYPE_CARD, 1);
> > 	card* pcard = *(card**) lua_touserdata(L, 1);
> > 	if(!lua_isnil(L, 2)) {
> > 		check_param(L, PARAM_TYPE_GROUP, 2);
> > 		group* pgroup = *(group**) lua_touserdata(L, 2);
> > 		pcard->set_material(&pgroup->container);
> > 	} else
> > 		pcard->set_material(0);
> > 	return 0;
> > }
> > ```
> > 
> > 
> > It should work properly with both the parameter being nil and with no parameters at all
> 
> in fact, the setmaterial(nil) works properly, but the `EFFECT_MATERIAL_CHECK` works incorrect when the parameter is nil. So we can change it to setmaterial(Group.CreateGroup()) or fix the `EFFECT_MATERIAL_CHECK`.

then the change should be made in set material, making it always call `EFFECT_MATERIAL_CHECK`, rather than changing it on the scripts

mallu11:
> > > Are you sure about set material not working with nil?
> > > ```
> > > int32 scriptlib::card_set_material(lua_State *L) {
> > > 	check_param_count(L, 2);
> > > 	check_param(L, PARAM_TYPE_CARD, 1);
> > > 	card* pcard = *(card**) lua_touserdata(L, 1);
> > > 	if(!lua_isnil(L, 2)) {
> > > 		check_param(L, PARAM_TYPE_GROUP, 2);
> > > 		group* pgroup = *(group**) lua_touserdata(L, 2);
> > > 		pcard->set_material(&pgroup->container);
> > > 	} else
> > > 		pcard->set_material(0);
> > > 	return 0;
> > > }
> > > ```
> > > 
> > > 
> > > It should work properly with both the parameter being nil and with no parameters at all
> > 
> > 
> > in fact, the setmaterial(nil) works properly, but the `EFFECT_MATERIAL_CHECK` works incorrect when the parameter is nil. So we can change it to setmaterial(Group.CreateGroup()) or fix the `EFFECT_MATERIAL_CHECK`.
> 
> then the change should be made in set material, making it always call `EFFECT_MATERIAL_CHECK`, rather than changing it on the scripts. Or change the script of the lyrilusc, so that it won't allow you to trigger the effect.

well, changing the setmaterial to make it always call `EFFECT_MATERIAL_CHECK` is better than changing the script of lyrilusc.

PR #1447, opened 2020-07-11
update プランキッズ・ウェザー for プランキッズ・ミュー

mallu11:
https://github.com/Fluorohydride/ygopro-pre-script/pull/486
fix: if there are only プランキッズ・ミュー and Rocksies in the GY, Weather Washer can banish プランキッズ・ミュー to activate it's effect and it won't be able to select 2 targets(after the banishment, only Rocksies left in the GY).


mallu11:
via https://github.com/Fluorohydride/ygopro-pre-script/pull/487

PR #1452, opened 2020-07-19
fix セフィラ・メタトロン

mallu11:


mallu11:
> `RESET_PHASE+PHASE_END` is necessary

the effect will be reset in condition if there not exist matching cards.

DailyShana:
see 亜空間物質転送装置
if End Phase is skipped, they will never return

PR #1453, opened 2020-07-20
Update c16317140.lua

LordOfNightmares:
The current code for adding tohand/spsummon leads to a weird scenario if in case you have 2 uria in ur gy where 1 is special summonable and the other isn't; 
When it asks you to select 1 of those cards you can select the undesired effect.

### DIFF
--- a/c16317140.lua
+++ b/c16317140.lua
@@ -77,28 +77,56 @@ function c16317140.atkop(e,tp,eg,ep,ev,re,r,rp)
 	end
 end
 function c16317140.spfilter(c,e,tp)
-	return c:IsCode(32491822,6007213,69890967)
-		and (c:IsAbleToHand() or (Duel.GetLocationCount(tp,LOCATION_MZONE)>0 and c:IsCanBeSpecialSummoned(e,0,tp,true,false)))
+	return c:IsCode(32491822,6007213,69890967) and Duel.GetLocationCount(tp,LOCATION_MZONE)>0 and c:IsCanBeSpecialSummoned(e,0,tp,true,false)
+end
+function c16317140.spfilter1(c)
+	return c:IsCode(32491822,6007213,69890967) and c:IsAbleToHand()
 end
 function c16317140.spcost(e,tp,eg,ep,ev,re,r,rp,chk)
 	if chk==0 then return Duel.IsExistingMatchingCard(Card.IsDiscardable,tp,LOCATION_HAND,0,1,nil) end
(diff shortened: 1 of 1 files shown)


mercury233:
If you want to special summon, but the activation is chained by _Vanity's Emptiness_, you should still add a card to hand.

In which case you cant distinguish which card can't be special summoned?

LordOfNightmares:
basically what happens when you have in your GY a Special summonable Uria and one that can only be added to hand is:
in `c16317140.spop` you select any of those 2 cards but you don't know what you selected for the summoning
then you check if it can be special summoned or added to hand so basically you might select the one which can be added to hand but cannot be special summoned in case where you want to special summon it
```lua
function c16317140.sptg(e,tp,eg,ep,ev,re,r,rp,chk)
	if chk==0 then return Duel.IsExistingMatchingCard(c16317140.spfilter,tp,LOCATION_GRAVE,0,1,nil,e,tp) end
	Duel.SetOperationInfo(0,CATEGORY_TOHAND,nil,0,tp,LOCATION_GRAVE)
	Duel.SetOperationInfo(0,CATEGORY_SPECIAL_SUMMON,nil,0,tp,LOCATION_GRAVE)
end
function c16317140.spop(e,tp,eg,ep,ev,re,r,rp)
	Duel.Hint(HINT_SELECTMSG,tp,HINTMSG_OPERATECARD)
	local sg=Duel.SelectMatchingCard(tp,aux.NecroValleyFilter(c16317140.spfilter),tp,LOCATION_GRAVE,0,1,1,nil,e,tp)
	local sc=sg:GetFirst()
	if sc then
		if Duel.GetLocationCount(tp,LOCATION_MZONE)>0 and sc:IsCanBeSpecialSummoned(e,0,tp,true,false)
			and (not sc:IsAbleToHand() or Duel.SelectOption(tp,1190,1152)==1) then
			Duel.SpecialSummon(sc,0,tp,tp,true,false,POS_FACEUP)
		else
			Duel.SendtoHand(sc,nil,REASON_EFFECT)
		end
	end
end
```

LordOfNightmares:
tldr; i addapted the code like for `Dingirsu, the Orcust of the Evening Star`

mallu11:
> tldr; i addapted the code like for `Dingirsu, the Orcust of the Evening Star`

无标题
无标题2
Dingirsu's effect has a selection when you activate the effect, but Hyper_Blaze is different, it's selection is in the solving of the effect.

LordOfNightmares:
> There is a sign to mark the uria is special summonable
oh yea u right nvm
image

PR #1457, opened 2020-07-29
triple tribute

salix5:
@DailyShana @mercury233 
Problem
1. オベリスクの巨神兵 will grant EFFECT_LIMIT_SET_PROC, and monster set is available.
2. 神獣王バルバロス will grant EFFECT_LIMIT_SUMMON_PROC, and "trbute 3 monsters to summon" is unavailable.

Solution
Now these monsters are classified into
--triple tribute(require 3 tributes, summon)
--triple tribute(require 3 tributes, set)
--triple tribute(can tribute 3 monsters, summon)
--triple tribute(can tribute 3 monsters, set)

~~Now all EFFECT_SUMMON_PROC with 3 tribute has summon type SUMMON_TYPE_ADVANCE+1.
The 1st effect of 守護神エクゾディア is changed into SUMMON_TYPE_ADVANCE+2.~~


DailyShana:
will implement `EFFECT_TRIPLE_TRIBUTE` properly, but not such complicated things

salix5:
@DailyShana @mercury233 
What is the current state of the triple tribute system?
Is it necessary to merge this one as a workaround?

mercury233:
I don't think this is a good approach...

salix5:
@DailyShana @mercury233 
What is the current state of the triple tribute system?
If it will not be ready in Spetember, can we merge this one as a workaround?

PR #1467, opened 2020-08-11
[Fix] Dragon Ravine 62265044

SoaringSky:
> Dragon Ravine
> Once per turn: You can discard 1 card, then activate 1 of these effects;
> ● Add 1 Level 4 or lower "Dragunity" monster from your Deck to your hand.
> ● Send 1 Dragon monster from your Deck to the GY.

**Issue:** The card script references string 2 and 3, which leads to prompt 1 being mislabeled, and prompt 2 being empty.

**Fix:** I edited SetDescription to reference the proper values.

**Example:**
Untitled


mercury233:
it had been there for years, so we should keep the string ids, or all sims will need to update their database

PR #1477, opened 2020-08-19
other prefer changes

purerosefallen:
1. Effects only for checking should always become EFFECT_FLAG_CANNOT_DISABLE.
2. Every cards sending this card to grave should all be checked.
3. **TODO** When special summoning other extra deck monsters sending this card to grave as a procedure, the effect resetting the flag would remain until next chain ends. Perhaps other reset effect with other timings are needed.


mallu11:
The reason card is not always the special summoned monster, here is an example:
Activating Tactical Exchanger to destroy 電脳堺姫-娘々 to special summon Autorokket Dragon. The reason card is a Trap Card, so should use other ways to fix it.

PR #1499, opened 2020-10-03
Fix Sentence of Doom

shadowfox87:
The flag was on the wrong card. The flag should be on Destiny Board not itself. This prevents Destiny Board from activating at the opponent's EP if this card is activated prior. Fixed by @Lyris12


mercury233:
The 2nd effect of this card won't affect the letter order of _Destiny Board_, and can be used even if no _Destiny Board_ is on field.

you should select L-A-N-I instead of I-N-A-L to speed up _Destiny Board_ with card.

Lyris12:
@mercury233 Unfortunately, that interaction is incorrect: https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=22691&keyword=&tag=-1&request_locale=ja

mercury233:
@Lyris12 what do you mean by the ruling? google translate say it mean _Sentence of Doom_ don't affect _Destiny Board_

Lyris12:
Oh. Sorry about that...

PR #1504, opened 2020-10-21
Fixed Token Sundae and Souten Calling 

TheOnePharaoh:
Source https://db.ygorganization.com/qa#270
"Tokens you control" includes Tokens in the spell and trap zone
still requires a core side fix to make sure type token is registered when tokens are in s/t zones


salix5:
トークン復活祭/Token Sundae
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=4&cid=9348
■自分のモンスターゾーンにモンスタートークンが存在しない場合には、「トークン復活祭」を発動する事はできません。
It actually means "Destroy all Tokens in Monster Zone".
It is possible that the descriptions of some old cards do not match its effect.

PR #1520, opened 2020-12-28
fix EFFECT_DISABLE_FIELD

salix5:
@mercury233 
EFFECT_FLAG_REPEAT is removed.
It is replaced by normal value function.
See  Fluorohydride/ygopro-core#351


salix5:
@mercury233 

PR #1521, opened 2021-01-10
update

KohanaSonogami5967:
it only indestructable one monster each turn
@mercury233


salix5:
Do you have a replay file that can display this bug?
Thank you in advance!

KohanaSonogami:
sorry I didn't test it, but some of my friend tells me that is it a bug. and i check all similarly of this effect so i change it.

PR #1536, opened 2021-01-28
Book of Lunar Eclipse by Soaring__Sky

SoaringSky:
I have no idea what I just did, maybe it works?


mercury233:
_it didn't actually work, and it is scripted earlier already_

SoaringSky:
I assumed  as much. Ah well, I gave it an earnest try

PR #1559, opened 2021-02-16
マドルチェ・シャトー must be interact with necrovalley

XyleN5967:
While the effect of my opponent's Necrovalley is applying, if I activate Madolche Chateau, which effects will be negated?
A:
The effect of Madolche Chateau that returns any "Madolche" monsters in your Graveyard to the Deck when it is first activated is an effect that is applied to a card(s) in the Graveyard, and that effect will be negated by Necrovalley, but Madolche Chateau will remain on the field and its effect that increases the ATK and DEF of all "Madolche" monsters on the field by 500 applies normally. Note that since the effects of "Madolche" monsters such as Madolche Mewfeuille that return them to the Deck when they are destroyed by the opponent's cards are effects that are applied to a card(s) in the Graveyard, and are negated by Necrovalley, the effect of Madolche Chateau that adds them to your hand instead cannot be applied as a result.

### DIFF
--- a/c14001430.lua
+++ b/c14001430.lua
@@ -39,7 +39,7 @@ function c14001430.target(e,tp,eg,ep,ev,re,r,rp,chk)
 end
 function c14001430.activate(e,tp,eg,ep,ev,re,r,rp)
 	if not e:GetHandler():IsRelateToEffect(e) then return end
-	local g=Duel.GetMatchingGroup(c14001430.tdfilter,tp,LOCATION_GRAVE,0,nil)
+	local g=Duel.GetMatchingGroup(aux.NecroValleyFilter(c14001430.tdfilter),tp,LOCATION_GRAVE,0,nil)
 	if g:GetCount()>0 then
 		Duel.SendtoDeck(g,nil,2,REASON_EFFECT)
 	end


mercury233:
https://github.com/Fluorohydride/ygopro-scripts/commit/46debde4528fc23700ed63e82dc86e92a3267e08

maybe those cards should create relation, to fix https://github.com/Fluorohydride/ygopro/issues/2022

XyleN5967:
@mercury233 i think not only cheteau

mercury233:
The problem is, if you chain _System Down_ to activate _Necrovalley_ (with _Metaverse_) when you got machine both in field and grave, should the monsters on field be banished or _System Down_ be negated entirely.

According to the text of _Necrovalley_, it should be negated, but there is [a ruling for _Rage with Eyes of Blue_](https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=12596&keyword=&tag=-1&request_locale=ja) saying the other cards will be banished as normal.

I think we should confirm whether this ruling is special only for _Rage with Eyes of Blue_.

shadowfox87:
There are similar rulings we can already use to determine the interaction between System Down vs Necrovalley.

> Q: While the effect of Necrovalley is applying, or if the effect of my opponent's Kycoo the Ghost Destroyer is applying, can I activate the effect of Trishula, Dragon of the Ice Barrier?

> A: In this scenario, due to the effect of your opponent's Kycoo the Ghost Destroyer (or Necrovalley), you cannot banish any cards from the Graveyard. Even in that scenario, you can activate the effect of Trishula, Dragon of the Ice Barrier, but you can only banish up to 2 of your opponent's cards (1 each from their hand and field.) Even if you banish Kycoo the Ghost Destroyer or Necrovalley on the field, you cannot banish any cards from the Graveyard.

Source: https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=9670&request_locale=ja

> Q: I activate System Down while there are Machine-Type monsters both on my opponent's side of the field and in their Graveyard. If my opponent chains Call of the Haunted and Special Summons Kycoo the Ghost Destroyer, how does the effect of System Down resolve?

> A: In this scenario, due to the effect of your opponent's Kycoo the Ghost Destroyer (or Necrovalley), you cannot banish any cards from the Graveyard. If there are any cards in your opponent's Graveyard, you cannot activate the effect of Nekroz of Sophia. If there are no cards in your opponent's Graveyard, you can activate the effect of Nekroz of Sophia. (The effect will be applied normally.)

Source: https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=7815&request_locale=ja

It is already clear from the above two rulings what should happen. The first ruling puts Necrovalley and Kycoo in the same category. They allow Trishula to resolve its effect, banishing cards from field and hand, but not GY.
`You can banish up to 1 card each from your opponent's hand, field, and GY. (The card in the hand is chosen at random.)` Remember however that Trishula's effect is optional for choosing to banish from each location. It's not a `and if you do` or requires banishment of cards in the GY for its effect to resolve.

This is why we needed to understand how System Down works because its effect states: `banish all Machine-Type monsters your opponent controls and in their Graveyard.` Here, the effect says "and in their GY" which means it is part of the effect and not optional. The second ruling allows System Down to resolve, banishing cards from the field only, but not in the GY.

Hence, based on logic, we can deduce what happens in this scenario:
> Q: Player 1 activates System Down while there is a face-up Machine-Type monster on the field and a Machine-Type monster in the Graveyard. Player 2 chains Metaverse and activates Necrovalley from their deck. What happens?

> A: System Down will resolve, banishing the face-up Machine Type monster on the field and not banish the monster in the GY.

shadowfox87:
https://db.ygorganization.com/qa#12596
> Q: Necrovalley is face-up in my Field Zone.

In this situation, can I activate Rage with Eyes of Blue?

> A: In this scenario, since the effect of Necrovalley prevents players from banishing cards from the Graveyard, you cannot activate Rage with Eyes of Blue if you have any cards in your Graveyard.

Additionally, if Metaverse is chained to Rage with Eyes of Blue and Necrovalley is activated from the Deck, you cannot banish any cards from the Graveyard but you still banish the cards on your field and in your hand, and Special Summon up to 3 Blue-Eyes White Dragons from your Deck.

Since Kycoo does not have the text that states, `Negate any card effect that would move a card in the Graveyard to a different place.`, a separate ruling is needed for Necrovalley. The above ruling is clear and proves that the effect will still resolve and not be negated. Only the monsters on the field and hand are banished, not GY.

PR #1574, opened 2021-02-24
fix センサー万別

mercury233:
the monsters which new appear on field must be select to be sent to grave


shadowfox87:
This is to fix this ruling so that if a face-up TCBOO is already on the field and a face-down monster is attacked, it will be sent immediately to Graveyard.

https://db.ygorganization.com/qa#9560

mallu11:
i think the turn player is the first one to send the monster to the grave

PR #1579, opened 2021-03-01
Fix "Revenge Rally"

Lyris12:
Its last effect always drew 1 card instead of 2 (with a discard) because the flag effect for that purpose was being reset beforehand.

### DIFF
--- a/c51686645.lua
+++ b/c51686645.lua
@@ -56,7 +56,7 @@ function c51686645.target(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
 	Duel.Hint(HINT_SELECTMSG,tp,HINTMSG_EQUIP)
 	Duel.SelectTarget(tp,Card.IsFaceup,tp,0,LOCATION_MZONE,1,1,nil)
 	Duel.SetOperationInfo(0,CATEGORY_EQUIP,e:GetHandler(),1,0,0)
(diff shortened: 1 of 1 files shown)


c51686645.lua
```diff
@@ -89,12 +89,15 @@ function c51686645.eqlimit(e,c)
 	return c:IsControler(1-e:GetHandlerPlayer())
 end
 function c51686645.drcon(e,tp,eg,ep,ev,re,r,rp)
-	return e:GetHandler():IsReason(REASON_LOST_TARGET)
-		and not e:GetHandler():GetPreviousEquipTarget():IsLocation(LOCATION_ONFIELD+LOCATION_OVERLAY)
+	if e:GetHandler():IsReason(REASON_LOST_TARGET)
+		and not e:GetHandler():GetPreviousEquipTarget():IsLocation(LOCATION_ONFIELD+LOCATION_OVERLAY) then return true end
+	e:GetHandler():ResetFlagEffect(51686645)
```
mercury233, line 94: why reset here?

Lyris12: 'Tis a security measure in case the card goes to the GY without its effect triggering, e.g. if the conditions for it aren't met.

mercury233: it will always be reset when the card leave grave and can trigger the 2nd effect again

Lyris12: Okay, but how many of the specified flag effect will `Card.ResetFlagEffect` remove?

PR #1585, opened 2021-03-07
fix 変導機咎 クロックアーク

salix5:
https://yugioh-wiki.net/index.php?%A5%DA%A5%F3%A5%C7%A5%E5%A5%E9%A5%E0%A5%B9%A5%B1%A1%BC%A5%EB
ペンデュラムスケールを表示する欄は２つ存在するが、左右の欄に異なるペンデュラムスケールが記されたカードは存在しない。
また、ルールブックにも「右のペンデュラムゾーンでは右のペンデュラムスケールを参照する」といった規定はない。
The rule book does not say that "The right Pendulum Scale is applied when the Pendulum Monster is put in the right zone."

https://www.yugioh-card.com/japan/howto/data/rulebook_masterrule20200401_ver1.0.pdf
マスタールール(2020年4月1日改訂版)対応公式ルールブック
ペンデュラムス召喚：P.43~P.44
When it mentions "Pendulum scale", it always uses the left and right number simultaneously.
The picture in P.44 also shows that the pendulum scale is a single number, no matter the Pendulum Monster is put in the left or right zone.

https://www.db.yugioh-card.com/yugiohdb/card_search.action
The form of official database takes Pendulum Scale as one value, just like Level/Rank.

Conclusion:
The evidence shows that the Pendulum Scale is one single number, and it is written twice in both sides for some reason.
~I think we can choose 1 fixed side (left or right), and all scripts keeps using that side.~

Edit:
The script should change both the left and right scale simultaneously.
When this effect is applied, its Pendulum Scale should be 1/1 (left zone) or 8/8 (right zone).


mercury233:
`c87468732.sccon` must be changed to differ left and right

salix5:
The condition function is fixed.

c87468732.lua
```diff
@@ -64,8 +70,11 @@ end
 function c87468732.cfilter(c)
 	return c:IsFacedown() or not c:IsType(TYPE_PENDULUM)
 end
-function c87468732.sccon(e)
-	return not Duel.IsExistingMatchingCard(c87468732.cfilter,e:GetHandlerPlayer(),LOCATION_MZONE,0,1,nil)
+function c87468732.sccon1(e)
+	return e:GetHandler():GetSequence()==0 and not Duel.IsExistingMatchingCard(c87468732.cfilter,e:GetHandlerPlayer(),LOCATION_MZONE,0,1,nil)
```
mercury233, line 74: `GetSequence` of PZONE will return 6 or 7 in MR3

salix5: In my opinion,  a card released in 2021 does not have to consider Master Rule 3, which ends in 2017.
We did not consider older rule like priority when Master Rule 3 was still active.

After Master Rule 3 ends, the whole rules have changed multiple times.
Many card descriptions have changed, and even the way to resolve some effects have also changed.
Keeping all scripts in the future compatible with Master Rule 3 is almost impossible, and it could make it difficult to fix the scripts.

If one really wants to play the game under Master Rule 3, one can fork this repo and revert it to 2017 Mar where Master Rule 3 is still active.
I think this repo itself should only take care of the current rules.

mercury233: it can be easily fixed by `c==Duel.GetFieldCard(tp,LOCATION_PZONE,0)`

mercury233 (approved):

PR #1606, opened 2021-03-31
fix センサー万別

mercury233:
fix: if the race of monsters are changed to the same, this card will still keep them

### DIFF
--- a/c24207889.lua
+++ b/c24207889.lua
@@ -51,14 +51,17 @@ function c24207889.adjustop(e,tp,eg,ep,ev,re,r,rp)
 	local phase=Duel.GetCurrentPhase()
 	if (phase==PHASE_DAMAGE and not Duel.IsDamageCalculated()) or phase==PHASE_DAMAGE_CAL then return end
 	local sg=Group.CreateGroup()
-	for p=0,1 do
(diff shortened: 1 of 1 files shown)


salix5:
i=0, 1, 2, 3
i=0, onlynew=true, p=0
It assumes that c24207889[p] is legal, and select cards based on c24207889[p].

i=1, onlynew=false, p=0
i=2, onlynew=true, p=1
i=3, onlynew=false, p=1

I think we should check c24207889[p] first to make sure that it is legal.

mercury233:
`c24207889[p]` is always a group, as set in global check

mercury233:
the problem is if _Reprodocus_ change a monster to dinosaur, should the player can select _Reprodocus_ to send to grave, or the new dinosaur must be sent to grave

salix5:
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=8763&keyword=&tag=-1
According to this question, I think the monster that is changed to new race should be sent to grave.

mercury233:
in that case we should record each race monster: https://github.com/Fluorohydride/ygopro-scripts/pull/1608

but i think we still need ruling

PR #1607, opened 2021-03-31
fix センサー万別

salix5:
@mercury233 
c24207889[p]: The legal set after the last adjustment.
sg: The card set that will be sent to grave in this adjustment.

Problem
The current script assumes that c24207889[p] is legal and selects cards based on c24207889[p].
This assumption could be wrong if c24207889[p] is changed into the same race by DNA改造手術.

Solution
in order to keep c24207889[p] legal, we should check c24207889[p] first.
Then we can use the original script to check other cards on the field.


mercury233:
did you see this? https://github.com/Fluorohydride/ygopro-scripts/pull/1606

PR #1614, opened 2021-04-04
Fixed War Rock Skyler and Bashileos

TheOnePharaoh:
Effect changes ATK as well so it should be able to activate in the damage step
https://db.ygorganization.com/card#15948


mercury233:
actually i don't think Skyler can special summon in damage step, but it seems cant be confirmed before ocg release

PR #1633, opened 2021-04-26
update ドラグニティアームズ－グラム

SmileEmoticon:


mercury233:
what will this fix?

SmileEmoticon:
it should be added since arma gram change the ATK stats of a monster along with disable effect

mercury233:
will it affect anything actually?

SmileEmoticon:
i am not sure

mercury233:
There do be a problem, if this card equip a monster by its 3rd effect and use its 2nd effect to disable itself, the equipped monster will be destroyed and it shouldn't count in the amount of equip cards to down the attack.
But `Duel.AdjustInstantly` won't fix this.

PR #1670, opened 2021-06-28
fix Duck Dummy

mercury233:
fix https://github.com/Fluorohydride/ygopro/issues/2355


mercury233:
_finally we had to use a similar solution like https://github.com/Fluorohydride/ygopro-scripts/pull/1483_

PR #1685, opened 2021-07-27
update continuous cards, part 1

salix5:
7/27
Split into 2 parts

# Problem
Continuous cards:
TYPE_CONTINUOUS Spell/Trap
TYPE_EQUIP, TYPE_FIELD

If a continuous card is not on the field, The effect resolution will do nothing.
This is the property of continuous cards, not the effect itself.

Suppose that a monster copies the effect of a continuous card, and it leaves the field before resolution.

Now:
The operation function will do nothing.

Correct:
The operation function will work as normal.

# Reason
The property is written in the card script.

Ex.
```lua
if not e:GetHandler():IsRelateToEffect(e) then return end
```

# Solution
For continuous cards, these conditions are unnecessary.
This property is checked by CHAIN_CONTINUOUS_CARD in ocgcore.

I removed the following conditions from continuous cards:
```lua
if not e:GetHandler():IsRelateToEffect(e) then return end
```

The list:
[cmd1.txt](https://github.com/Fluorohydride/ygopro-scripts/files/6882853/cmd1.txt)

Total: 431 cards

Modified scripts that are not continuous cards:
[cmd1.txt](https://github.com/Fluorohydride/ygopro-scripts/files/6882855/cmd1.txt)

Result:
15684835
21350571
43405287
49551909
63049052
97182396

These cards are Normal Trap cards which can become equip cards, so the conditions should be removed too.

--
8/3
[pendulum.txt](https://github.com/Fluorohydride/ygopro-scripts/files/6921255/pendulum.txt)
modified: 78 pendulum monsters
rebased to master

# Description of modified cards
[desc_cmd.zip](https://github.com/Fluorohydride/ygopro-scripts/files/6882948/desc_cmd.zip)
https://docs.google.com/spreadsheets/d/1WKGNdRtDdMWxDpEckAC56R-tX1klS_BDOKSyo5PVKfk/edit?usp=sharing


salix5:
@mercury233 
Orange color in the table:
Effects activated in some place other than LOCATION_SZONE which will move the card itself.

These effect has to check Card.IsRelateToEffect(), and they may be accidentally changed by the replacement.
I have checked all these cards, and デーモンの杖 is the only accidentally changed card.

salix5:
@mercury233 

mercury233:
```
if not c:IsRelateToEffect(e) then return end
```
are you planing to check this variant?

salix5:
Confirming

salix5:
> 
> 
> ```
> if not c:IsRelateToEffect(e) then return end
> ```
> 
> are you planing to check this variant?

#1694

PR #1695, opened 2021-08-22
Fix several Continuous Traps

Lyris12:
Was able to target itself in the same chain link it was activated

### DIFF
--- a/c13364097.lua
+++ b/c13364097.lua
@@ -38,9 +38,11 @@ end
 function c13364097.target(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
 	local g=Duel.GetMatchingGroup(c13364097.tdfilter,tp,LOCATION_REMOVED,0,nil)
 	if chkc then return chkc:IsOnField() and chkc:IsFaceup() end
-	if chk==0 then return Duel.IsExistingTarget(Card.IsFaceup,tp,LOCATION_ONFIELD,LOCATION_ONFIELD,1,nil) and g:CheckSubGroup(aux.dncheck,2,2) end
+	local xg=Group.CreateGroup()
+	if not e:GetHandler():IsStatus(STATUS_EFFECT_ENABLED) then xg:AddCard(e:GetHandler()) end
+	if chk==0 then return Duel.IsExistingTarget(Card.IsFaceup,tp,LOCATION_ONFIELD,LOCATION_ONFIELD,1,xg) and g:CheckSubGroup(aux.dncheck,2,2) end
 	Duel.Hint(HINT_SELECTMSG,tp,HINTMSG_DESTROY)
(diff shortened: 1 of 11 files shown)


salix5:
Filter functions will be called very often.
Creating a new group each time may result in redundant groups.

mercury233:
why not use the same way as _Unending Nightmare_

mercury233:
@salix5 but there is no other way to exclude 2 cards when checking the cost of _PSY-Frame Overload_

salix5:
I think there is a way to do so.
I am typing on mobile device and it may take several minutes.

salix5:
function c36970611.tgfilter(c,tp,xc)
return c:IsAbleToRemove(tp,POS_FACEDOWN) and c~=xc
end

function c36970611.cfilter(c,tp,xc)
	return c:IsSetCard(0xc1) and c:IsType(TYPE_MONSTER) and (c:IsLocation(LOCATION_HAND) or c:IsFaceup()) and c:IsAbleToRemoveAsCost()
		and Duel.IsExistingTarget(c36970611.tgfilter,tp,LOCATION_ONFIELD,LOCATION_ONFIELD,1,c,tp,xc)
end

function c36970611.cost(e,tp,eg,ep,ev,re,r,rp,chk)
	local xc=nil
	if not e:GetHandler():IsStatus(STATUS_EFFECT_ENABLED) then xc=e:GetHandler() end
	if chk==0 then return Duel.IsExistingMatchingCard(c36970611.cfilter,tp,LOCATION_HAND+LOCATION_MZONE,0,1,nil,tp,xc) end
	Duel.Hint(HINT_SELECTMSG,tp,HINTMSG_REMOVE)
	local cg=Duel.SelectMatchingCard(tp,c36970611.cfilter,tp,LOCATION_HAND+LOCATION_MZONE,0,1,1,nil,tp,xc)
	Duel.Remove(cg,POS_FACEUP,REASON_COST)
end
function c36970611.target(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
	if chkc then return chkc:IsOnField() and chkc:IsAbleToRemove() end
	if chk==0 then return true end
	local xc=nil
	if not e:GetHandler():IsStatus(STATUS_EFFECT_ENABLED) then xc=e:GetHandler() end
	Duel.Hint(HINT_SELECTMSG,tp,HINTMSG_REMOVE)
	local g=Duel.SelectTarget(tp,c36970611.tgfilter,tp,LOCATION_ONFIELD,LOCATION_ONFIELD,1,1,nil,tp,xc)
	Duel.SetOperationInfo(0,CATEGORY_REMOVE,g,1,0,0)
end

Lyris12:
> @Lyris12 why not use the same way as _Unending Nightmare_

Yeeah good point; done

PR #1697, opened 2021-08-24
fix 機巧嘴－八咫御先, 交差する魂, 天空聖騎士アークパーシアス

Sonic714:
Fixed an issue with 機巧嘴－八咫御先 that special summon limit does not take effect when the monster normal summoned is destroyed by continuous effects(e.g. Burning Abyss' self-destruction), by adjusting its reset methods.
Fixed 交差する魂 for the same issue as in #1679 .
Fixed an issue with 天空聖騎士アークパーシアス that it can banish face-down monsters as cost.


salix5:
機巧嘴－八咫御先
fixed

交差する魂
confirming

PR #1707, opened 2021-09-21
fix activation count

salix5:
@mercury233 
# Problem
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=20232&keyword=&tag=-1&request_locale=ja
The effect activation is counted when the monster effect is negated.

There are 2 types of counting:
1. oath type
If the activation is negated, it is NOT treated as "activate 1 effect".

2. normal type
If the activation is negated, it is still treated as "activate 1 effect".

https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=22342&keyword=&tag=-1&request_locale=ja
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=23158&keyword=&tag=-1&request_locale=ja
The count of 運命のドロー, 交差する魂 is "oath type".

# Solution
運命のドロー
交差する魂
The global counter of oath type can be replaced by 
Duel.AddCustomActivityCounter(id,ACTIVITY_CHAIN,chainfilter)

chainfilter
The reverse filter.
The counter will count the number of effects that do not match this filter.


salix5:
> Because there is an issue with card activation effects like Pop-Up that in latest ruling it should be not able to activate when cards limiting activation count like Great Shogun Shien and 交差する魂 take effect. I am troubled by this issue for months and cannot think of any solution.

Do you have a reference for this ruling?
Thank you!

Sonic714:
> > Because there is an issue with card activation effects like Pop-Up that in latest ruling it should be not able to activate when cards limiting activation count like Great Shogun Shien and 交差する魂 take effect. I am troubled by this issue for months and cannot think of any solution.
> 
> Do you have a reference for this ruling? Thank you!

I'm sorry for absence because of being busy recently. I attempt to add a structure like summon limit counter(but more detailed) to handle this problem.
Q.
1、相手のフィールドに「大将軍 紫炎」が1枚あります。自分のデッキには「魔鐘洞」が1枚あります。この場合、自分は「ポップルアップ」や「終焉の地」を発動できますか？ 
A.
ご質問の場合、「ポップルアップ」や「終焉の地」を発動することはできません。

PR #1708, opened 2021-09-24
add GetID()

salix5:
The function proposed in  Fluorohydride/ygopro#2367 .

### DIFF
--- a/utility.lua
+++ b/utility.lua
@@ -4,6 +4,15 @@ POS_FACEUP_DEFENCE=POS_FACEUP_DEFENSE
 POS_FACEDOWN_DEFENCE=POS_FACEDOWN_DEFENSE
 RACE_CYBERS=RACE_CYBERSE
 
+function GetID()
+	local str=string.match(debug.getinfo(2,"S")["source"],"c%d+%.lua")
+	str=string.sub(str,1,string.len(str)-4)
+	local scard=_G[str]
+	local id=tonumber(string.sub(str,2))
+	local offset=id<100000000 and 1 or 100
+	return scard,id,offset
+end
+
 --the lua version of the bit32 lib, which is deprecated in lua 5.3
 bit={}
 function bit.band(a,b)


salix5:
@mercury233 
Can we merge this PR?

Yuzurisa:
By nature of `debug` lib it's very unsafe because it can access Lua internal registry, causing security problems. It has potential of hacking the server OS from card scripts.
I believe in future YGOPro updates, in terms of security, this lib should be DISABLED. See https://github.com/Fluorohydride/ygopro/issues/2378 for details.
As a conclusion, we should avoid using those dangerous libs in official scripts.

pyrQ:
> By nature of `debug` lib it's very unsafe because it can access Lua internal registry, causing security problems. It has potential of hacking the server OS from card scripts. I believe in future YGOPro updates, in terms of security, this lib should be DISABLED. See [Fluorohydride/ygopro#2378](https://github.com/Fluorohydride/ygopro/issues/2378) for details. As a conclusion, we should avoid using those dangerous libs in official scripts.

Alternatively you can just look at the scripts you're adding to your server. Not to mention that you can do things like eating up all the RAM without any weird libs being used at all.

Lyris12:
Or, even better, why not instead use `Card.GetOriginalCode`?

edo9300:
@Lyris12 this function is called when the lua script itself is called, you have no Card object or any other core element available. So you don't have anything to call that function on.

Lyris12:
> @Lyris12 this function is called when the lua script itself is called, you have no Card object or any other core element available. So you don't have anything to call that function on.

I know; I was thinking 1 add a parameter to the `GetID` function for use by the `Card.GetOriginalCode` function, 2 declare the card metatable, ID, and offset outside all functions in the card script, but without initializing them until the card's `initial_effect` function is called, at which point the variables are initialized.
```
local s,id,o
function cID.initial_effect(c)
    s,id,o=GetID(c)
end
```... I guess because y'all still want only 1 place, if any, to be modified when a card's ID changes?

**Comment by edo9300** [comment]
> > @Lyris12 this function is called when the lua script itself is called, you have no Card object or any other core element available. So you don't have anything to call that function on.
> 
> I know; I was thinking 1 add a parameter to the `GetID` function for use by the `Card.GetOriginalCode` function, 2 declare the card metatable, ID, and offset outside all functions in the card script, but without initializing them until the card's `initial_effect` function is called, at which point the variables are initialized.
> 
> ```
> local s,id,o
> function cID.initial_effect(c)
>     s,id,o=GetID(c)
> end
> ```
> 
> ... I guess because y'all still want only 1 place, if any, to be modified when a card's ID changes?

The purpose of this function is to be able to retrieve that ``cID`` and the other values in a simple and easy way for the scripter, what you're proposing is just an overly complex solution (not always working also) for an already solved problem. Also as I already said in the other issue, this is an implementation that works out of the box without modifications of the core, with a small addittion of the core that GetID function can be integrated directly into it and made it return more accurate values (not relying on the filename) without having to change any other script that already started using that format.

**Comment by Lyris12** [comment]
Yeeaa okay...

**Comment by purerosefallen** [comment]
It's not worthy to introduce a security problem as a drawback of scripting convenience.

**Comment by purerosefallen** [comment]
@pyrQ The measure of bad things are different.
If you do something like eating all memory, the worst thing could be ygopro crashes and quits.
However, remote code execution could lead to data leak or loss, destroying the whole server.

**Comment by pyrQ** [comment]
> Due to the nature of how custom card community works, anyone could submit a card, using the vulnerable of debug library, and hack the server easily.

If you blindly accept code that will be executed on your server without looking at it then that's your problem.

**Comment by edo9300** [comment]
@purerosefallen the debug library could be harmful when used with harmful intent, in this case it's a single function call being used in a "standard" function, there's really no possibility to abuse that function call. Also nothing is preventing to add a simple
```
debug=nil
```
Immediately afterwards to just remove any possibility to use functions afterwards.

PR #1721, opened 2021-10-28
update Duel.SendToDeck

mercury233:
Use constant
Always use "shuffle" instead of "top" when a shuffle will be followed, to work with pseudo shuffle mode


salix5:
confirming

mercury233:
@salix5 I had checked the replacement of constants

PR #1728, opened 2021-11-06
update Rebirth of Nephthys

chronogenexx:
use `aux.AddRitualProcGreater2` and `extra_operation`.


utility.lua
```diff
@@ -1697,7 +1697,7 @@ function Auxiliary.RitualUltimateOperation(filter,level_function,greater_or_equa
 					tc:CompleteProcedure()
 				end
 				if extra_operation then
-					extra_operation(e,tp,eg,ep,ev,re,r,rp,tc)
+					extra_operation(e,tp,eg,ep,ev,re,r,rp,tc,mat)
```
mercury233, line 1700: `mat` is a local var which is not in this scope

chronogenexx: https://github.com/Fluorohydride/ygopro-scripts/pull/1728/commits/27f4bf5fc722ad58e3d1ed542742106ba67b29a8 fixed.

PR #1747, opened 2021-11-17
remove unnecessary tab/space

chronogenexx:


purerosefallen:
OK for KPI, but thanks anyway.

PR #1764, opened 2021-12-04
Fix Terror of Trishula, Predaplant Moray Nepenthes, Infinitrack Xyz

Sonic714:
Fixed Predaplant Moray Nepenthes and Infinitrack Xyz monsters and now they can only capture GY, face-up banished and face-up Ex monsters.
Fixed Terror of Trishula and now it can activate as long as any card can be banished under the requirements of Syncro monsters.


mercury233:
please place different fixes in different pull request

PR #1767, opened 2021-12-05
fix シンデレラ

Lyris12:
Redtext occurs on line 53 while there are no available targets for her 1st effect to equip


mercury233:
any replay? it is already checked by `Duel.IsExistingMatchingCard`

Lyris12:
> any replay? it is already checked by `Duel.IsExistingMatchingCard`

For example, respond to her effect with `Needlebug Nest` while the target(s) it's among the last 5 cards in the user's deck

PR #1780, opened 2021-12-28
fix パーペチュアルキングデーモン

Lyris12:
The last effect now triggers only once for Fiend monsters which are sent to the same player's GY sequentially, such as by the Ignition Effect of a `Dark Grepher` and/or by multiple activated effects a single same Chain which would send a Fiend(s) to the user's GY.

Reason: In the situation listed above and others like it, players currently have a hard time distinguishing which Fiend will leave the GY with Masterking's last effect, since all instances of it read as follows: `Roll dice`


mercury233:
maybe we should add some flag to add all event groups to the event in the core

PR #1785, opened 2022-01-01
part fix イタチの大暴発

mercury233:
If one monster's ATK is above 0xFFFF(65535), `MSG_SELECT_SUM` won't work because `opParam` has the max of 0xFFFFFFFF and will be splited.

### DIFF
--- a/c31044787.lua
+++ b/c31044787.lua
@@ -25,10 +25,16 @@ function c31044787.target(e,tp,eg,ep,ev,re,r,rp,chk)
 end
 function c31044787.activate(e,tp,eg,ep,ev,re,r,rp)
 	local g=Duel.GetMatchingGroup(c31044787.filter,tp,0,LOCATION_MZONE,nil,1-tp)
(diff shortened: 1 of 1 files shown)


salix5:
Maybe we should modify `MSG_SELECT_SUM`, `PROCESSOR_SELECT_SUM`.
`min` and `max` should be treated as int32.

salix5:
@mercury233 
https://github.com/Fluorohydride/ygopro-core/blob/169d577bc8cfd133b54c9d829b540626245537ac/playerop.cpp#L632
```cpp
pduel->write_buffer32(acc & 0xffff);
```

Why do we use bitwise and here?

mercury233:
the other operation params can't get beyond 0xffff

purerosefallen:
How about, lower the precision of attack count for 1/10.
Like, we may treat atk 1000 as value 100, and treat LP 8000 as LP 800. In this way number would be much smaller that monster with atk 655350 could fit.

PR #1790, opened 2022-01-11
fix EFFECT_SWAP_BASE_AD

salix5:
# Problem
Swap ATK and DEF:
assign ATK = current DEF
assign DEF = current ATK

Swap base ATK and DEF should be the same


salix5:
@mercury233 
These are all of them.
Sorry for any inconvenience I have brought.

salix5:
我也不想老是改這個
所以我要把它一次改好

PR #1812, opened 2022-02-12
Fix Lib the World Key Blademaster

Sonic714:
Fixed an issure that World Key Blademaster still limited effect activation in the case of the card had already activated.


salix5:
fixed in 88ff1fb8cf85294dd528d1ba26c2f91d38ccb5d7
Thank you!

PR #1822, opened 2022-03-02
Fix メルフィータイム

Lyris12:
Was forced to detach as many Xyz Materials as possible as activation cost


mercury233:
`Duel.SelectMatchingCard` can select 0 card for `sg` too

Lyris12:
> `Duel.SelectMatchingCard` can select 0 card for `sg` too

Is choosing not to select anything handled client-side, or core-side? https://cdn.discordapp.com/attachments/253948767630393344/948454043808849960/unknown.png

mercury233:
image
I guess it is client side

PR #1826, opened 2022-03-08
add aux.xyz_number

salix5:
@mercury233 

# Problem 
https://github.com/Fluorohydride/ygopro/issues/2425

# Reason
The `xyz_number` is in a table which is outside the game.

# Solution
Some effects can change the card name into a card name which is outside the game.
Therefore, we should store the properties related to card names in a global table.

Now `xyz_number` is stored in
```lua
Auxiliary.xyz_number[id]
```

No. with a number: 147 (except No. XX)
竜装合体 ドラゴニック・ホープレイ/Ultimate Dragonic Utopia Ray
獣装合体 ライオ・ホープレイ/Ultimate Leo Utopia Ray

total: 149

2 of them are changed in the first commit (Number 78, Number 99: Utopia Dragonar).
147 of them are changed in the second commit.


mercury233:
shouldn't we put them in each scripts?

salix5:
The main problem is:
If A change its name into B (id=456) and B is not in the deck, c456.lua is not loaded.

So I think we should put the table in a common script (constant or utility).

mercury233:
we can put both
```
Auxiliary.xyz_number[68679595]=39
Auxiliary.xyz_number[56840427]=39
```
in c68679595.lua

mercury233:
we should add
```lua
function Auxiliary.GetXyzNumber(v)
	local id
	if Auxiliary.GetValueType(v)=="Card" then id=v:GetCode() end
	if Auxiliary.GetValueType(v)=="number" then id=v end
	return Auxiliary.xyz_number[id]
end
```
to replace the `_G` way to get number

salix5:
> we can put both
> 
> ```
> Auxiliary.xyz_number[68679595]=39
> Auxiliary.xyz_number[56840427]=39
> ```
> 
> in c68679595.lua

Agree



> we should add
> 
> ```lua
> function Auxiliary.GetXyzNumber(v)
> 	local id
> 	if Auxiliary.GetValueType(v)=="Card" then id=v:GetCode() end
> 	if Auxiliary.GetValueType(v)=="number" then id=v end
> 	return Auxiliary.xyz_number[id]
> end
> ```
> 
> to replace the `_G` way to get number

Why don't we look up the table `Auxiliary.xyz_number` directly?

mercury233:
`aux.GetXyzNumber(c)` is shorter and easier to use than `aux.xyz_number[c:GetCode()]`

PR #1830, opened 2022-03-11
add aux.GetCappedValue()

salix5:
@mercury233 

# Problem
Like #1785, the value is wrong when lv > 0xffff

# Solution
The value is capped at 0xffff.


mercury233:
do any card can make the level go beyond 65535 without loop?

salix5:
It might happen in certain cases, and it can be prevented easily.
So I think we should fix it.

PR #1841, opened 2022-03-21
fix EFFECT_CANNOT_TRIGGER

salix5:
@mercury233 
1. remove unnecessary `SetValue()`
2. EFFECT_CANNOT_TRIGGER with a target should use forcibly registration.


salix5:
3/23
update part 2
remove string: str4 of 59975920

c59975920.lua
```diff
@@ -71,10 +71,8 @@ function c59975920.spop(e,tp,eg,ep,ev,re,r,rp)
 	local tc=Duel.GetFirstTarget()
 	if tc:IsRelateToEffect(e) and Duel.SpecialSummonStep(tc,0,tp,tp,false,false,POS_FACEUP) then
 		local e1=Effect.CreateEffect(e:GetHandler())
-		e1:SetDescription(aux.Stringid(59975920,3))
```
mercury233, line 74: why remove this

salix5: There are a lot of effects doing things like "Special Summon X, the effect of X cannot be activated".
Most of them does not have client hints.

mercury233: I think we should add hints to them instead of remove.
https://github.com/Fluorohydride/ygopro-scripts/pull/409/files#diff-a17888fba9bcfaa5abb726d6f0730fd45ea19869480b0f4ee327e7fea11baa25

salix5: OK
revert to previous version

PR #1869, opened 2022-04-28
fix "Negate the effect, destroy that card" effects

mercury233:
If the activating card is disabled already, those effects shouldn't destroy them.
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=12361&keyword=&tag=-1&request_locale=ja

Fix:
If _Greedy Venom Fusion Dragon_ activated the 2nd effect, and chained by _Starlight Road_, special summon _Greedy Venom Fusion Dragon_ back to the field and let it be negated by _Skill Drain_, the effect of _Starlight Road_ should still resolve.


mercury233:
img

PR #1878, opened 2022-05-07
fix Purushaddoll Aeon, Satellarknight Rigel

465uytrewq:
fix: if target monster was unaffected by the effects, target monster change face-down position in the end phase.
> https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=20087&keyword=&tag=-1
> なお、『対象のモンスターは攻撃力・守備力が１０００アップし、エンドフェイズに裏側守備表示になる』処理は、**対象のモンスターが罠カードの効果を受けず、適用されません**。


mercury233:
I think the ruling means if the monster is not affected by the atk gain effect, it shouldn't turn upside down at the end phase.

chronogenexx:
we should fix Satellarknight Rigel?

mercury233:
> should we fix Satellarknight Rigel?

I think so

chronogenexx:
https://github.com/Fluorohydride/ygopro-scripts/pull/1878/commits/eeac4b1dd7eeee9e1c527578d9bd5f8ed8e9635b fixed.

salix5:
695a21becb4b04512b47e61e08d6d614cea5116d
Thank you!

PR #1893, opened 2022-06-02
fix Earthbound Immortal Uru and so on

chronogenexx:
fix: if player's main monster zone's space don't exist while Lair of Darkness has on the field, Earthbound Immortal Uru can release opponent monster.


c59255742.lua
```diff
@@ -80,13 +80,17 @@ function c59255742.spop(e,tp,eg,ep,ev,re,r,rp)
 		Duel.SpecialSummon(tc,0,tp,tp,false,false,POS_FACEUP)
 	end
 end
+function c59255742.costfilter(c,tp,g)
```
mercury233, line 83: `c59255742.costfilter` isn't used

mercury233 (changes requested):

c59255742.lua
chronogenexx: https://github.com/Fluorohydride/ygopro-scripts/pull/1893/commits/0f0d1b2be73eebe69f0af7d3d0573fb10ae268e1 fixed.

PR #1903, opened 2022-06-15
clean abuse of Card.IsRelateToEffect

Yuzurisa:
We in the past had abused `Card.IsRelateToEffect` so much. In fact relations are established to chains rather than effects. This mistake actually caused issues like https://github.com/Fluorohydride/ygopro/issues/2389 . It's time to clean them up.


c20537097.lua
```diff
@@ -59,12 +59,12 @@ function c20537097.thtg(e,tp,eg,ep,ev,re,r,rp,chk)
 	local g=Duel.GetMatchingGroup(c20537097.thfilter,tp,LOCATION_DECK,0,nil)
 	local ct=c:GetColumnGroup():FilterCount(Card.IsControler,nil,1-tp)
 	if c:IsControler(1-tp) then ct=ct+1 end
-	if chk==0 then return c:IsRelateToEffect(e) and ct>0 and g:GetClassCount(Card.GetCode)>=ct end
+	if chk==0 then return c:IsRelateToChain(0) and ct>0 and g:GetClassCount(Card.GetCode)>=ct end
```
mercury233, line 62: won't return true here

mercury233 (changes requested):

PR #1923, opened 2022-07-23
Temp Revert "Fix Nine Pillars of Yang Zing (#1922)"

salix5:
@mercury233 @Sonic714 
This reverts commit 50a70cd5dbeed8bd530b74717dd9b87b8e3a20e0.
Revert #1922 temporarily.

# Reason
There are some evidences that the original version might be correct.
(Ex. The situation in Master Duel, the mail in 7/23)

Therefore, I think we should revert to the original version until further confirmation.


Sonic714:
I think reverting is ok.

PR #1931, opened 2022-07-31
fix Evil Mind, Dogmatika Encounter and so on

465uytrewq:
fix: Serial Spell, Trap of Darkness, ラビュリンス・バラージュ and so on don't check 2 or more effect.

> https://yugioh-wiki.net/index.php?%A1%D4%B5%A1%B9%C3%C9%F4%C2%E2%A4%CE%BA%C6%CA%D4%C0%A9%A1%D5#faq
> Ｑ：《機甲部隊の再編制》の[発動](https://yugioh-wiki.net/index.php?%C8%AF%C6%B0)に対して[《連続魔法》](https://yugioh-wiki.net/index.php?%A1%D4%CF%A2%C2%B3%CB%E2%CB%A1%A1%D5)を[発動](https://yugioh-wiki.net/index.php?%C8%AF%C6%B0)する事はできますか？
> Ａ：[《連続魔法》](https://yugioh-wiki.net/index.php?%A1%D4%CF%A2%C2%B3%CB%E2%CB%A1%A1%D5)を[チェーン](https://yugioh-wiki.net/index.php?%A5%C1%A5%A7%A1%BC%A5%F3)して[発動](https://yugioh-wiki.net/index.php?%C8%AF%C6%B0)でき、**その際に2つの『●』の[効果](https://yugioh-wiki.net/index.php?%B8%FA%B2%CC)から、1つを選択します**。
> 　　なお、その際に選択した[効果](https://yugioh-wiki.net/index.php?%B8%FA%B2%CC)の[コスト](https://yugioh-wiki.net/index.php?%A5%B3%A5%B9%A5%C8)となる、[手札](https://yugioh-wiki.net/index.php?%BC%EA%BB%A5)の任意の[カード](https://yugioh-wiki.net/index.php?%A5%AB%A1%BC%A5%C9)や「[マシンナーズ](https://yugioh-wiki.net/index.php?%A5%DE%A5%B7%A5%F3%A5%CA%A1%BC%A5%BA)」と名のついた[カード](https://yugioh-wiki.net/index.php?%A5%AB%A1%BC%A5%C9)は捨てません。(20/10/15)

PR #1934, opened 2022-08-09
add Effect.SetCostCheck

chronogenexx:
need https://github.com/Fluorohydride/ygopro-core/pull/460.


mercury233:
what will this fix, without https://github.com/Fluorohydride/ygopro-scripts/pull/1931 ?

chronogenexx:
`Effect.IsCostChecked` always return true because these scripts don't use `Card.CheckActivateEffect`.
so, `Effect.IsCostChecked` can return false if this fix.

mercury233 (approved):

PR #1681, opened 2021-07-21
fix & update pendulum summon for Pendulum Encore

465uytrewq:
https://github.com/Fluorohydride/ygopro-core/pull/390

PR #1952, opened 2022-08-30
Update c6992184.lua

Aephiex:
Issue: May select pendulum ritual monsters from your hand, but nothing happens. They aren't pendulum summoned and they stay in your hand.

Workaround: Makes it ignore summon condition & ignore revive limit when it tries to actually special summon the monsters. You cannot still pendulum summon monsters you couldn't before, because they cannot be selected in aux.PendOperation().


chronogenexx:
https://github.com/Fluorohydride/ygopro-scripts/pull/1681

PR #1955, opened 2022-09-02
fix 鎖龍蛇－スカルデット

salix5:
fix https://github.com/Fluorohydride/ygopro/issues/2441


mercury233:
Cards like バウンドリンク, 聖邪のステンドグラス may need fix too

salix5:
Auxiliary.PlaceOnBottom()
add
```lua
Duel.ShuffleHand(p)
```
Now the opponent cannot know which cards are selected.

mercury233:
I think we should rename this function

salix5:
Any suggestions?

mercury233 (changes requested):

utility.lua
```diff
@@ -2984,3 +2984,30 @@ function Auxiliary.ThisCardInGraveAlreadyReset2(e)
...
 	e:Reset()
 end
+--Player p place n cards from his hand on the bottom of his Deck in any order
+function Auxiliary.PlaceOnBottom(p,n)
+	local g=Duel.GetMatchingGroup(Card.IsAbleToDeck,p,LOCATION_HAND,0,nil)
+	if g:GetCount()<n then return end
+	Duel.ShuffleHand(p)
+	Duel.BreakEffect()
+	Duel.Hint(HINT_SELECTMSG,p,HINTMSG_TODECK)
+	local sg=g:Select(p,n,n,nil)
```
mercury233, line 2994: We should support _Dark Spirit's Mastery_ by allow passing `sg` as param

mercury233: or another function to simply Tearlaments scripts

salix5: So the function should be something like:
```lua
function Auxiliary.PlaceGroupOnDeckBottom(p,sg)
end
```

The script should generate sg by themselves.
Is it right?

mercury233: I think so

salix5:
Should we merge this one?

utility.lua
```diff
@@ -2984,3 +2984,26 @@ function Auxiliary.ThisCardInGraveAlreadyReset2(e)
...
+		end
+	end
+	if x2>0 then
+		Duel.SortDecktop(p,1-p,x2)
+		for i=1,x2 do
+			local mg=Duel.GetDecktopGroup(1-p,1)
+			Duel.MoveSequence(mg:GetFirst(),SEQ_DECKBOTTOM)
+		end
+	end
+	return #sg2
```
mercury233, line 3008: why only return sg2

mallu11:
@salix5 why this return #og? If there are some effects that use `xxx for each card returned`, I think this return the operation group is better (both deck and extra deck).

mercury233:
> @salix5 why this return #og? If there are some effects that use `xxx for each card returned`, I think this return the operation group is better (both deck and extra deck).

Because other similar functions such as `Duel.Destroy`, `Duel.Remove`, `Duel.SendtoGrave` return the operated amount.

mallu11:
> > @salix5 why this return #og? If there are some effects that use `xxx for each card returned`, I think this return the operation group is better (both deck and extra deck).
> 
> Because other similar functions such as `Duel.Destroy`, `Duel.Remove`, `Duel.SendtoGrave` return the operated amount.

I see. But #og is the operated amount of Main Deck, it should include the returned card(s) in Extra Deck.

PR #1981, opened 2022-10-11
update 召喚師アレイスター

mercury233:
There is an alternative art of _Invocation_


mercury233:
Should we update _The White Stone of Legend_ and _Malefic Blue-Eyes White Dragon_ and so on?

PR #1989, opened 2022-11-03
fix センサー万別

salix5:
@mercury233 
# Problem
unknown
[bug_only_one.zip](https://github.com/Fluorohydride/ygopro-scripts/files/9930567/bug_only_one.zip)
Turn 3
The first `There Can Be Only One` is destroyed, and then the second `There Can Be Only One` is activated.
When the effect resolves, the player cannot choose `Number 4: Stealth Kragen` in Main Monster Zone.

# Reason
The groups should be cleared when `There Can Be Only One` is not applied.

# Solution
EFFECT_RACE_ONLY_ONE:
the mark of the effects:
> Each player can only control 1 monster of each Type. 

Ex. `There Can Be Only One` 

If all EFFECT_RACE_ONLY_ONE are removed, the groups should be cleared.

### DIFF
--- a/c24207889.lua
+++ b/c24207889.lua
@@ -6,13 +6,13 @@ function c24207889.initial_effect(c)
 	e1:SetCode(EVENT_FREE_CHAIN)
 	e1:SetHintTiming(0,TIMINGS_CHECK_MONSTER)
 	c:RegisterEffect(e1)
-	--adjust
+	--only one
 	local e2=Effect.CreateEffect(c)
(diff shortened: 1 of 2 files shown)


mercury233:
~~Blue-Eyes Spirit Dragon need effect constant more~~

salix5:
> ~Blue-Eyes Spirit Dragon need effect constant more~

Maybe we can add it later.

mercury233:
I still don't think we should use an effect code only for 1 script

salix5:
done

PR #1992, opened 2022-11-10
fix King Beast Barbaros and so on

465uytrewq:
fix: if target card(s) are only equip card, _King Beast Barbaros_ can activate effect.
(but ygopro sent an error message.)

### DIFF
--- a/c11516241.lua
+++ b/c11516241.lua
@@ -46,8 +46,11 @@ end
 function c11516241.efilter(e,te)
 	return te:GetOwner()~=e:GetOwner()
 end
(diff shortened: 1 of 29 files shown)


mercury233:
Thank you!
I'm trying to find a solution for more scenes in https://github.com/Fluorohydride/ygopro/issues/2343

465uytrewq:
It would be better to merge https://github.com/Fluorohydride/ygopro-scripts/pull/1995 than this commit.
So, this commit is closing.

PR #1563, opened 2021-02-17
fix Necrovalley and Ghost Belle

mercury233:
fix https://github.com/Fluorohydride/ygopro-scripts/pull/1559

For effects that target cards in grave with definite action (eg. _Monster Reborn_):
- `SetOperationInfo` with the target, without location
- Necrovalley negate them in Necrovalley's script, by checking `tg:IsExists`
- Ghost Belle negate them by checking `tg:IsExists`

For effects that target cards in grave but with not definite action (eg. _World Legacy Guardragon_):
- SetOperationInfo without the part of grave
- `SetCategory` with `CATEGORY_GRAVE_ACTION` or `CATEGORY_GRAVE_SPSUMMON`
- Use `aux.NecroValleyNegateCheck(tg)` in the script of the effect
- Ghost Belle negate them by checking `IsHasCategory(CATEGORY_GRAVE_ACTION)`

For effects that don't target but only affect 1 card in grave and won't resolve without card in grave (eg. _Salamangreat Sunlight Wolf_):
- `SetOperationInfo` with `nil` as target, with `count`>0 and `param`==`LOCATION_GRAVE` exactly
- Necrovalley negate them in Necrovalley's script, by checking `v==LOCATION_GRAVE and ct>0`
- Ghost Belle negate them by checking `dv&LOCATION_GRAVE==LOCATION_GRAVE`

For effects that don't target but affect cards in grave and won't resolve without card in grave (eg. _Lazion, the Timelord_):
- `SetOperationInfo` with affected cards as target, with `count`>0
- Use `aux.NecroValleyNegateCheck(g)` in the script of the effect
- Ghost Belle negate them by checking `tg:IsExists`

For effects that don't target but affect cards in grave and may resolve without card in grave (eg. _Preparation of Rites_):
- `SetOperationInfo` without the part of grave
- `SetCategory` with `CATEGORY_GRAVE_ACTION` or `CATEGORY_GRAVE_SPSUMMON`
- Use `aux.NecroValleyFilter` in the script of the effect
- Ghost Belle negate them by checking `IsHasCategory(CATEGORY_GRAVE_ACTION)`

For effects that don't target but may affect cards in grave and may resolve without card in grave (eg. _Miracle Fusion_):
- `SetOperationInfo` with `nil` as target, with `count`>0 and `param` including `LOCATION_GRAVE`
- `SetCategory` don't need `CATEGORY_GRAVE_ACTION` or `CATEGORY_GRAVE_SPSUMMON`
[...]

PR #2002, opened 2022-12-10
fix Elemental HERO Necroid Shaman

chronogenexx:
fix: when _Elemental HERO Necroid Shaman_'s effect activated, _Ghost Belle & Haunted Mansion_ cannot activate.


salix5:
fixed in 7bfd325ff98d70cd47215d74ebbe5fc579475654

According to #1563 
> For effects that don't target but only affect 1 card in grave and won't resolve without card in grave (eg. _Salamangreat Sunlight Wolf_):
> 
>     * `SetOperationInfo` with `nil` as target, with `count`>0 and `param`==`LOCATION_GRAVE` exactly
> 
>     * Necrovalley negate them in Necrovalley's script, by checking `v==LOCATION_GRAVE and ct>0`
> 
>     * Ghost Belle negate them by checking `dv&LOCATION_GRAVE==LOCATION_GRAVE`
> 
> 
> For effects that don't target but affect cards in grave and won't resolve without card in grave (eg. _Lazion, the Timelord_):
> 
>     * `SetOperationInfo` with affected cards as target, with `count`>0
> 
>     * Use `aux.NecroValleyNegateCheck(g)` in the script of the effect
> 
>     * Ghost Belle negate them by checking `tg:IsExists`
Mandatory parts must move cards in grave:
It is negated by Necrovalley.

PR #2021, opened 2023-02-12
update 百檎龍－リンゴブルム and more sync monsters

mercury233:
@salix5


salix5 (approved):

salix5:
Changed: 百檎龍－リンゴブルム
I think the token should not have a continuous effect.

mercury233:
but all `EFFECT_NONTUNER` use `EFFECT_FLAG_SINGLE_RANGE`

salix5:
> but all `EFFECT_NONTUNER` use `EFFECT_FLAG_SINGLE_RANGE`

My opinion:
Like ATK change, `EFFECT_NONTUNER` or `EFFECT_TUNER` can be:
- temporary status
- continuous effect 

I think a token or a normal monster should not have a continuous effect, so I prefer temporary status.

PR #2022, opened 2023-02-13
fix 工作箱 1

Lyris12:


Lyris12:
Cards picked by the user, but not the opponent, were not put on the bottom of the Deck.

PR #2026, opened 2023-02-25
fix 工作箱 2

Lyris12:
Cards picked by the user, but not the opponent, were put on the bottom of the Deck, but their Deck then gets immediately shuffled.


c70508653.lua
```diff
@@ -35,10 +35,10 @@ function c70508653.thop(e,tp,eg,ep,ev,re,r,rp)
 		Duel.Hint(HINT_SELECTMSG,tp,HINTMSG_CONFIRM)
 		local sg=g:SelectSubGroup(tp,aux.dncheck,false,2,2)
 		Duel.ConfirmCards(1-tp,sg)
-		Duel.ShuffleDeck(tp)
 		local tg=sg:RandomSelect(1-tp,1)
 		tg:GetFirst():SetStatus(STATUS_TO_HAND_WITHOUT_CONFIRM,true)
 		Duel.SendtoHand(tg,nil,REASON_EFFECT)
```
Lyris12, line 40: Alternatively:
```suggestion
		Duel.DisableShuffleCheck()
		Duel.SendtoHand(tg,nil,REASON_EFFECT)
```

PR #2048, opened 2023-03-11
syn summon using SynMix can be canceled

mercury233:


salix5:
Should we avoid goto?

mercury233:
The main problem of `goto` is it may make the code harder to read. But in this approach it just lead the way to the start of selecting material procedure so it won't be problem here.
The another problem is it may cause unexpected context. But lua has the "jumps into the scope of local" check aleady. And the local vars are reinitialized after goto.

salix5 (approved):

PR #2053, opened 2023-03-16
fix ドラグニティ－ジャベリン

salix5:
# Problem
https://github.com/Fluorohydride/ygopro-core/pull/492

# Solution
Now it use `EFFECT_DESTROY_REPLACE` instead.


mercury233:
The effect of Javelin is, after it is destroyed, it can be equiped, but it is still treated as destroyed. `EFFECT_DESTROY_REPLACE` won't treat the card as destroyed.

PR #2060, opened 2023-03-23
replace EFFECT_CANNOT_ATTACK_ANNOUNCE with EFFECT_CANNOT_ATTACK

chronogenexx:
these effect is not "monsters cannot declare an attack", but "monsters cannot attack".

> https://www.db.yugioh-card.com/yugiohdb/card_search.action?ope=2&cid=5979
> このカードを発動するターン、自分の「青眼の白龍」は**攻撃できない**。


mercury233:
`EFFECT_CANNOT_ATTACK` won't apply to player

mercury233:
The can't attack limit of _Evolution Burst_ etc. shouldn't be treated as effect, because all other "このカードを発動するターン" texts aren't treated as effect, so it should has 'EFFECT_FLAG_IGNORE_IMMUNE'.

chronogenexx:
https://github.com/Fluorohydride/ygopro-scripts/pull/2060/commits/edc44f16f0f4c1ff01cbbed038baaa82ae7ae828 fixed.

PR #2066, opened 2023-03-28
fix Destructive Daruma Karma Cannon

Lyris12:
https://forum.duelistsunite.org/t/daruma-karma-cannon-bug/16108


mercury233:
refer: https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=21666&keyword=&tag=-1&request_locale=ja

PR #2068, opened 2023-04-01
update ThisCardInGraveAlreadyCheck & fix Assault Synchron

mercury233:
There is a ruling that if _Assault Synchron_ and _Chaos Ruler, the Chaotic Magical Dragon_ (summoned by its 2nd effect) are used as synchro material, _Assault Synchron_ can't use its effect in grave because it is not in grave when Chaos Ruler is banished.

Formerly, `Auxiliary.AddThisCardInGraveAlreadyCheck` can't be used at this case because those `EVENT_REMOVE` and `EVENT_TO_GRAVE` events are checked together, so the operation of the ThisCardInGraveAlreadyCheck will be called after the condition of `EVENT_REMOVE`. We can move the operation to condition to fix this.

The problem of the formerly implement of that ruling in _Assault Synchron_ is, if it was used as material of a synchro monster, and the synchro monster was returned to extra deck, and the same synchro monster is summoned again using Chaos Ruler, it should be able to use effect, but `c:GetReasonEffect()` is same at this condition. So we had to reset this reg in ThisCardInGraveAlreadyCheck.

### DIFF
--- a/c77202120.lua
+++ b/c77202120.lua
@@ -57,16 +57,14 @@ end
 function c77202120.splimit(e,c)
 	return not c:IsType(TYPE_SYNCHRO) and c:IsLocation(LOCATION_EXTRA)
 end
-function c77202120.cfilter(c,e,tp,se,cre)
+function c77202120.cfilter(c,tp,se)
 	return c:IsControler(tp) and c:IsRace(RACE_DRAGON) and c:IsType(TYPE_SYNCHRO)
 		and c:IsPreviousLocation(LOCATION_ONFIELD) and c:IsPreviousControler(tp)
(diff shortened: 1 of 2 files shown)


chronogenexx:
if _Assault Synchron_ and dragon-type synchro monster were tributed by _Raviel, Lord of Phantasms - Shimmering Scraper_'s sumon, _Assault Synchron_ can activate.
(because of _Raviel, Lord of Phantasms - Shimmering Scraper_ is not used `REASON_MATERIAL`, but `REASON_COST`.)

mercury233:
`EFFECT_SPSUMMON_PROC` is fixed, but the normal Tribute Summon is not fixed

PR #2073, opened 2023-04-10
fix `Bystial Dis Pater`

Lyris12:
was unable to revive opposing banished LIGHT/DARK monsters


mercury233:
that part is in `c27572350.distg`

Lyris12:
oh; I was looking at the wrong function

PR #2080, opened 2023-05-05
fix シューティング・クェーサー・ドラゴン

salix5:
If the monster will grant an effect according to the number or condition of material, we can add the effect in `EFFECT_MATERIAL_CHECK` directly.

Reason:
1. Effects which change the stat of a monster should be added before it enters the field.
2. Not using `e:SetLabel()` and `e:GetLabel()` will simplify the control flow of the script.

Test:
[test_material.zip](https://github.com/Fluorohydride/ygopro-scripts/files/11406640/test_material.zip)
test_material.lua
Single mode test script

### DIFF
--- a/c35952884.lua
+++ b/c35952884.lua
@@ -11,17 +11,10 @@ function c35952884.initial_effect(c)
 	e1:SetValue(aux.synlimit)
 	c:RegisterEffect(e1)
 	--multi attack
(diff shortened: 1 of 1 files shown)


mercury233:
But it can be copied by _Supreme King Dragon Starving Venom_

PR #2094, opened 2023-07-23
fix cards

Lyris12:
`Nazuki the Vaylantz Ninja`
* Try targeting an opposing monster with the monster effect.
`TGX3-DX2`
* There now exist "T.G." Spells & Traps.
`Dyna Mondo`
* String ID inconsistency

### DIFF
--- a/c3868277.lua
+++ b/c3868277.lua
@@ -11,7 +11,7 @@ function c3868277.initial_effect(c)
 	c:RegisterEffect(e1)
 end
 function c3868277.filter(c)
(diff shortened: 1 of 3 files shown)


salix5:
Thank you very much!
minor fix: 1d481dca8ffd291fcd4394c048b4280d7c040caf

https://github.com/Fluorohydride/ygopro-core/commit/74de21327d696aa1460fc228da6267ed456c1110
Duel.SelectField()
from the viewpoint of playerid
lower byte: self
upper byte: opponent

PR #2100, opened 2023-08-12
fix `Baelgrill de Nouvelles`

Lyris12:
Tributing monsters was mandatory

https://forum.duelistsunite.org/t/baelgrill-de-nouvelles-bug/17308


Lyris12:
https://yugioh-diffs.github.io/#18744

PR #2101, opened 2023-08-14
add .gitignore

salix5:
@mercury233 
~global.lua
It contains the global tables defined in ocgcore.
The Debug library is not included because it should not appear in normal script.
It will not be loaded in ygopro.~

~By adding global.lua, we can:~
~1. Use the Lua extension in VS Code to detect syntax errors.~
~2. Check if the script file name marches the card id in cards.cdb.~

The card table is generated by
```js
SELECT datas.id FROM datas, texts WHERE datas.id == texts.id 
AND abs(datas.id - alias) >= 10 
AND NOT type & ${TYPE_TOKEN} 
AND (NOT type & ${TYPE_NORMAL} OR type & ${TYPE_PENDULUM})
```
~


mercury233:
Which extension are you using?

salix5:
I don't have my computer at hand.
I searched Lua and tried the one with most downloaded count.
I will check it later.

salix5:
We can add `.gitignore` first.

mercury233:
https://github.com/mercury233/ygopro-scripts-define

the sketch, use with `Lua.workspace.library` setting

PR #2154, opened 2023-10-23
fix Auxiliary.GCheckAdditional() call

salix5:
@mercury233 
The search result of `GCheckAdditional=` (Match whole word):
This function has only 2 versions:
function(g)
function(g, ec)

It seems that the 3rd, 4th, ... parameters are unnecessary.

Last change:
#1269


salix5:
fixed in d0a226a4ac029dcfe7945e7abe6f70fee301bbff

PR #2175, opened 2023-10-31
fix-融合解除

Crescent0v0:
fix: you always can't respond this card with 屋敷わらし Ghost Belle & Haunted Mansion 
****
修复：不能被屋敷童对应的问题


mercury233:
reverted in https://github.com/Fluorohydride/ygopro-scripts/commit/e45d2d75808fdb8576d32e4fa39c80ec865951c5
`CATEGORY_GRAVE_SPSUMMON` will do this.

PR #2258, opened 2023-12-03
Update Guardian's code list for プリンシパグ

Wind2009-Louse:


c34904525.lua
```diff
@@ -1,8 +1,8 @@
 --雷風魔神－ゲート・ガーディアン
 function c34904525.initial_effect(c)
+	aux.AddCodeList(c,25955164,62340868)
```
465uytrewq, line 3: `aux.AddFusionProcMix` and `aux.AddFusionProcMixRep` is calling `aux.AddCodeList` by default.

PR #2281, opened 2023-12-13
Fix 幻惑の巻物

Wind2009-Louse:
When 幻惑の巻物 is equipped by other card's effect, the equip monster will lost attribute.

幻惑之卷物用其它卡的效果装备时，会让装备怪兽失去属性。


salix5:
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=9002&keyword=&tag=-1&request_locale=ja

PR #2290, opened 2023-12-21
update & simplify aux.FusionProcMix

465uytrewq:
update1: if fusion monster has many materials(e.g. _Ultimate Crystal Rainbow Dragon Overdrive_), ygopro is running slower.
so, I optimized `aux.FusionProcMix`.
update2: `aux.AddFusionProcMixRep` merge with `aux.AddFusionProcMix`.
update3: remove unused `mg` parameter.
update4: add `gcheck` parameter(deal with 究極竜魔導師).


purerosefallen:
What exactly is the improvement of the algorithmic complexity?

Could you tell me that, what it looks like before, and after?

procedure.lua
```diff
@@ -1051,149 +932,136 @@ function Auxiliary.FOperationMixRep(insf,sub,fun1,minc,maxc,...)
 				local concat_fusion=chkfnf&0x200>0
 				local sub2=(sub or notfusion) and not concat_fusion
 				local mg=eg:Filter(Auxiliary.FConditionFilterMix,c,c,sub2,concat_fusion,fun1,table.unpack(funs))
+				::FOperationMixSelectStart::
```
purerosefallen, line 935: Please DON'T use goto label.

mercury233: I think we can use `goto` reasonably.

purerosefallen:
Let's talk about this one again.

Could we, DON'T change the format of the mat filter, preserving mg?

PR #2303, opened 2023-12-26
Fix Noble Arms

vuryleo:
fix Noble Arms, it should check the target card (monster) instead of card (equipment)


salix5:
Thank you!

PR #2322, opened 2024-01-05
fix 超融合

salix5:
@mercury233 
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=23510&keyword=&tag=-1&request_locale=ja
We have to check if the fusion material is affected by the effect.

Test replay:
[test_fusion.zip](https://github.com/Fluorohydride/ygopro-scripts/files/13841858/test_fusion.zip)


mercury233:
https://github.com/Fluorohydride/ygopro-scripts/pull/1635

PR #2331, opened 2024-01-08
fix aux.mzctcheckrel

465uytrewq:
@salix5 

fix: if 聖霊獣騎 レイラウタリ has on the field, _Ten Thousand Dragon_ cannot special summon.


salix5:
Thank you!

PR #2347, opened 2024-01-15
fix `ゴーティスの朧キーフ`

Lyris12:
User was allowed to control no Fish monsters for it to activate its self-Summoning effect, while the opponent controlled any

see also https://github.com/mycard/ygopro-scripts/pull/28

### DIFF
--- a/c99529628.lua
+++ b/c99529628.lua
@@ -49,7 +49,7 @@ function s.cfilter(c)
 	return c:IsFaceup() and c:IsRace(RACE_FISH)
 end
 function s.spcon(e,tp,eg,ep,ev,re,r,rp)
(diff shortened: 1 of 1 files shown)


mercury233:
> ①：フィールドに魚族モンスターが存在する場合に発動できる。このカードを手札から特殊召喚する。
https://www.db.yugioh-card.com/yugiohdb/card_search.action?ope=2&cid=19470&request_locale=ja

It doesn't require the Fish monster is on some player's field

mercury233:
yugipedia had fixed
https://yugipedia.com/index.php?title=Kiif,_Haze_of_the_Ghoti&diff=prev&oldid=4978081

PR #2348, opened 2024-01-15
fix Auxiliary.SameValueCheck

mercury233:


mallu11:
we should add different value check too, if use `g:GetClassCount(Card.GetLinkAttribute)==#g`, **Dark Simorgh** and **Red-Eyes Black Dragon** can link summon **Ib the World Chalice Priestess**.

PR #2359, opened 2024-01-17
add Auxiliary.SelectDeckCard

salix5:
@mercury233 
@purerosefallen 
Edit: update to e4dc825

```lua
---Select cards from multiple locations including Field or Deck
---@param tp integer
---@param filter function|nil
---@param player integer
---@param location_s integer
---@param location_o integer
---@param min integer
---@param max integer
---@param ex Card|Group|nil
---@param ... any
---@return Group
---@return boolean
function Auxiliary.SelectMultipleLocationCard(tp,filter,player,location_s,location_o,min,max,ex,...) end

---Select target from multiple locations including Field
---@param tp integer
---@param filter function|nil
---@param player integer
---@param location_s integer
---@param location_o integer
---@param min integer
---@param max integer
---@param ex Card|Group|nil
---@param ... any
---@return Group
function Auxiliary.SelectMultipleLocationTarget(tp,filter,player,location_s,location_o,min,max,ex,...) end
```


purerosefallen:
An improved version of it as at branch `patch-select-deck-2`.

It asks the player to click on a card cover to reveal the deck, and added another return value telling if the player has viewed the deck.

salix5:
Thank you.

mercury233:
The parameters of the new function should be exact same with `Duel.SelectMatchingCard`, we may need a function to read the hint, and a `Duel.SelectOption` which don't tell the opponent which option you selected.

salix5:
@mercury233 
done

salix5:
@mercury233 
@purerosefallen 
What should we do to this PR
- merge this PR as a workaround
- close as not planned
- other

mercury233:
waiting for better solution

PR #2385, opened 2024-01-27
fix--GP－Nブラスター

kiritosoft:
fix GP－Nブラスター will destroy field spell ( Controller is GP－Nブラスター's  Controller) when its effect target monster which Sequence is 3.


kiritosoft:
fix GP－Nブラスター's effect target monster in extra monster zone and destroy it but it can not destroy extra card (you can destroy all cards in its adjacent Monster Zones and Spell & Trap Zones.) when LP is lower than opponent

kiritosoft:
fix GP－Nブラスター's effect can destroy Non Monster Card. 
fix GP－Nブラスター's effect will destroy wrong card when target card was changed controller

PR #2405, opened 2024-02-14
limit value Duel.AnnounceNumber

465uytrewq:
~~need to https://github.com/Fluorohydride/ygopro-core/pull/561.~~


salix5:
confirming
I am not sure if it is safe to pass 10000+ arguments to a Lua function.

PR #2418, opened 2024-02-19
Update c49930315.lua

VforViolin:
Fix SetCountLimit wrong ID


mercury233:
It is preferred to use different id to make it clear that the effect don't share limit with another effect

VforViolin:
Ahh understood

PR #2419, opened 2024-02-19
Update c49930315.lua

VforViolin:
Mark it as a card that can treat itself as a tuner, as per card text.


salix5:
57712bd2f58d8c6a47df66c156d58bc447944ba3
Merged into master.
Thank you.

PR #2420, opened 2024-02-21
Fix 転臨の守護竜

Crescent0v0:
Add exMaterial monsters isTypeLink Check 
(exMaterial monsters from effect like [チェーン・マテリアル
Chain Material](https://www.db.yugioh-card.com/yugiohdb/card_search.action?ope=2&cid=7466&request_locale=ja))


salix5:
@mercury233 
Can you check the fusion summon part?
Thank you.

PR #2440, opened 2024-03-06
Fix destroyed by an opponent's card (by battle or card effect)

kiritosoft:
修复持有（这张卡被对方破坏的场合才能发动。）的效果的卡和相同攻击力的对手衍生物进行战斗且双方都被战斗破坏的场合无法发动效果。

fix effect(If this card in its owner's possession is destroyed by an opponent's card (by battle or card effect):) can not activate when they  destroyed by battle with same attack token monster.(token and that monster were  all battle destroyed)


salix5:
96a2ee26fe2ecbed09186fb9e33dc1590b41dfe5
EVENT_BATTLE_DESTROYED'
rp改成戰鬥時的控制者而非傷害步驟結束時的控制者

PR #2452, opened 2024-03-20
fix `Twin-Barrel Dragon`

Lyris12:
It was allowed to activate its effect if the opponent had nothing for it to target


mercury233:
It is mandatory effect so it will activate without target

PR #2460, opened 2024-03-25
Fix shuffle into deck on resolution

VforViolin:
Previous commit (eeac7bf) broke the effect to shuffle back into the deck.


mercury233:
https://github.com/Fluorohydride/ygopro-core/pull/555 is required to work with that commit, did you test it?

PR #2477, opened 2024-04-03
Fix Moon Mirror Shield cost

VforViolin:
MMS should pick where in the deck it returns to on resolution, not on activation of it's trigger effect.


salix5:
fixed in 41c885de9f877190417b9cf065080c30f2620286
Thank you!

PR #2480, opened 2024-04-04
add EFFECT_FLAG_ACTIVATE_CONDITION

mercury233:
see https://github.com/Fluorohydride/ygopro-core/pull/571


mercury233:
the CNo.101 and 103 change require https://github.com/Fluorohydride/ygopro-scripts/pull/2500

PR #2493, opened 2024-04-15
Add aux.AddCodeList() for field spells

Wind2009-Louse:
For Krishnerd Witch


salix5:
https://github.com/Fluorohydride/ygopro-core/pull/579
We can use this function in the script of Krishnerd Witch.

PR #2502, opened 2024-04-19
support card metatable for aux.AddCodeList

purerosefallen:
Now we can do things like this in the bottom of script.

```lua
aux.AddCodeList(c10000000, 10000010)
```

For effects like Krishnerd Witch, I'm planning to auto-manage this mention list using some automatic tools, so this is necessary.


purerosefallen:
ping

PR #2519, opened 2024-05-07
fix: `Necrovalley` + `G.B. Hunter` vs Pendulum

Lyris12:
Effects like that of `Majespecter Draco - Ryu` should be negated if any attempt is made to move it from the GY to a different place, but a loophole was present where Ryu _can_ set itself from the GY into a Pendulum Scale while `Necrovalley` is active.

This should _not_ be the case.


Lyris12:
related to https://github.com/Fluorohydride/ygopro/issues/2543

PR #2530, opened 2024-05-14
fix--どぐう

kiritosoft:
fix どぐう's effect don't have count limit


mercury233:
https://github.com/Fluorohydride/ygopro-scripts/commit/3b174c34fd4706c56fcd2ea05dab0765bd678809

PR #2432, opened 2024-02-28
updates for Duel.PreserveSelectDeckSequence

purerosefallen:
requires https://github.com/Fluorohydride/ygopro-core/pull/564

PR #2541, opened 2024-05-28
add aux.SelectPDS and fixed related cards

purerosefallen:
image

fixed this.


mercury233 (changes requested):
significantly harder to read

purerosefallen:
> significantly harder to read

I tried something like `aux.SelectPreservingDeckSequence` but it's a  pretty long name

mercury233:
https://github.com/Fluorohydride/ygopro-scripts/pull/2432#issuecomment-2094047878

purerosefallen:
@mercury233 how about `aux.SelectShowDeck`?

mercury233:
it is obviously unacceptable to add such wrap, please use appropriate method

purerosefallen:
> it is obviously unacceptable to add such wrap, please use appropriate method

so any better ways?

mercury233:
```
Duel.ShowDeckSequence()
local sg=g:Select(...)
Duel.ShowDeckSequence(false)
```
nobody will know what is P or what is Preserve

purerosefallen:
Are you arguing about the naming problem, or you want to have a brand new way to solve this problem?

Please think of ways to solve the problem, rather than closing other's PR or deleting other's branch.

purerosefallen:
a

purerosefallen:
> ```
> Duel.ShowDeckSequence()
> local sg=g:Select(...)
> Duel.ShowDeckSequence(false)
> ```
> 
> nobody will know what is P or what is Preserve

Considering the number of cards should be changed, we have to do something on aux.

If you could think of a better name, you could do a direct replacement on those files.

mercury233:
> > ```
> > Duel.ShowDeckSequence()
> > local sg=g:Select(...)
> > Duel.ShowDeckSequence(false)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > nobody will know what is P or what is Preserve
> 
> Considering the number of cards should be changed, we have to do something on aux.
> 
> If you could think of a better name, you could do a direct replacement on those files.

Please change "we" to "I" (you)

purerosefallen:
> > > ```
> > > Duel.ShowDeckSequence()
> > > local sg=g:Select(...)
> > > Duel.ShowDeckSequence(false)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > nobody will know what is P or what is Preserve
> > 
> > 
> > Considering the number of cards should be changed, we have to do something on aux.
> > If you could think of a better name, you could do a direct replacement on those files.
> 
> Please change "we" to "I" (you)

Let's consider something real. Which name would you like to use rather than the currenct suffix `PDS`?

Or if you want another way, could you introduce it?

mercury233:
https://github.com/Fluorohydride/ygopro-scripts/pull/2432#issuecomment-2094047878

purerosefallen:
> [#2432 (comment)](https://github.com/Fluorohydride/ygopro-scripts/pull/2432#issuecomment-2094047878)

How about in THIS PR?

I think something like `aux.SelectShowDeckSequence` but it's still pretty long.

mercury233:
I think it is better to do something to remove `Duel.ConfirmCards` from those effects and close https://github.com/Fluorohydride/ygopro/pull/2470
Anyway you shouldn't (and can't) add a wrap version to every select function, and `aux.selpds` will be even better than `aux.SelectPDS` because the latter looks like a new type of abbreviation

purerosefallen:
> I think it is better to do something to remove `Duel.ConfirmCards` from those effects and close [Fluorohydride/ygopro#2470](https://github.com/Fluorohydride/ygopro/pull/2470) Anyway you shouldn't (and can't) add a wrap version to every select function, and `aux.selpds` will be even better than `aux.SelectPDS` because the latter looks like a new type of abbreviation

1. Previous time I asked if you would merge https://github.com/Fluorohydride/ygopro/pull/2470 on the incoming release, but I got a reply like *no time to do ygopro2*.
2. This function is only needed to select cards from deck when it's on deck top or bottom, or opponent selecting deck cards, being a rare case. In addition, the idea prior to https://github.com/Fluorohydride/ygopro-core/pull/564 which is adding an `aux.SelectDeckCard` you also supported the idea to wrap every select method, and that one is a much wider case. So what exactly is your idea on wrapping select method?
3. Or should we do them in the core way instead? I'm thinking of hiding deck sequence in selecting player's own decks only, showing deck sequence as usual when selecting opponent's deck.

purerosefallen:
https://github.com/Fluorohydride/ygopro-core/pull/595

PR #2547, opened 2024-06-01
fix スピードリフト

salix5:
#2546 
Maybe we should avoid reset manually.


mercury233:
I doubt whether this will work

salix5:
[test#2547.zip](https://github.com/user-attachments/files/15524797/test.2547.zip)
C1 スピードリフト 
QE timing at chain end:
The opponent cannot activate maxx C

C1 Normal Spell
QE timing at chain end:
The opponent can activate maxx C

mercury233:
_didn't check it yet_

PR #2556, opened 2024-06-07
Update c78783557.lua

VforViolin:
Fix Veidos so it can be summoned to the opponent's field by it's own effect, while only you control a Pyro monster, and are under the effects of TCBOO.


salix5:
For some reasons, the card after certain pack is not on the master branch.
Can you compare your branch to `origin/PHNI-EN`?
Thank you very much!

VforViolin:
Done: https://github.com/Fluorohydride/ygopro-scripts/pull/2559

PR #2562, opened 2024-06-08
Update c32909498.lua

VforViolin:
Fixed so the effect does not trigger if it changes control before the chain in which the trigger effect would meet its activation conditions.


salix5:
2a6ce7741e2e2530d25d4ee9cff710bf80e661a2

PR #2573, opened 2024-06-14
Add auxiliary.register merged delayed event to single card

kiritosoft:
Auxiliary.RegisterMergedDelayedEvent have some problem about triggering effect.
for example:
1.Zombie World in field zone, use Mezuki's effect to special summon Kashtira Arise-Heart. And that Kashtira Arise-Heart can activate effect to Attach Mezuki which was banished to this card as material. 
2.The Bystial Lubellion place Branded Regained in chain 1 and Bystial Magnamhut remove monster in chain 2. Branded Regained can activate effect(If a LIGHT or DARK monster(s) is banished (except during the Damage Step): You can target 1 of those monsters; place that monster on the bottom of the Deck, and if you do, draw 1 card.).

They should not activate that effect at that time.

so, Add a new function to solve that problem。


YiLong-jin:
This function was written by me. But I entrusted kiritosoft to submit it on my behalf. If you have any questions about this change, you can leave me a comment here.
The idea behind this change is to rule that cards that require the disclosure area to be exposed and have not gone through the full trigger process cannot activate the inducing effect. The old code didn't distinguish between each card, so it caused bugs. The new program I wrote was to use separate events for each card to trigger the timing separately.
In the test, the scenario test requirements of the person responsible for adjudication were passed before the upload was made.

YiLong-jin:
The corpse said that it was okay to consider overwriting the old functions。I'll start a new PR in a while。

utility.lua
```diff
@@ -1480,6 +1480,80 @@ function Auxiliary.MergedDelayEventCheck2(e,tp,eg,ep,ev,re,r,rp)
...
+	e1:SetRange(0xff)
+	e1:SetLabel(event_code_singel)
+	e1:SetLabelObject(g)
+	e1:SetOperation(Auxiliary.MergedDelayEventCheck1_ToSingelCard)
+	c:RegisterEffect(e1)
+	local e2=e1:Clone()
+	e2:SetCode(EVENT_CHAIN_END)
+	e2:SetOperation(Auxiliary.MergedDelayEventCheck2_ToSingelCard)
+	c:RegisterEffect(e2)
+	--当卡片移动到公开区域时，重置监测卡片组
```
purerosefallen, line 1508: Please write comment in English

purerosefallen (changes requested):

c1688285.lua
```diff
@@ -25,19 +25,19 @@ function s.initial_effect(c)
 	e2:SetOperation(s.thop)
 	c:RegisterEffect(e2)
 	--material
+	local g=Group.CreateGroup()
+	local custom_code=aux.RegisterMergedDelayedEvent_ToSingelCard(c,id,EVENT_TO_GRAVE,g)
+	aux.RegisterMergedDelayedEvent_ToSingelCard(c,id,EVENT_REMOVE,g,custom_code)
```
purerosefallen, line 30: Is it a spelling mistake? This is OCG card and please avoid them

utility.lua
```diff
@@ -1480,6 +1480,80 @@ function Auxiliary.MergedDelayEventCheck2(e,tp,eg,ep,ev,re,r,rp)
 		g:Clear()
 	end
 end
+--Once the card has been moved to the public area, it should be listened to again
+function Auxiliary.RegisterMergedDelayedEvent_ToSingelCard(c,code,event,g,event_code_singel)
+	local mt=getmetatable(c)
+	if not event_code_singel then 
+		event_code_singel=code+event+EVENT_CUSTOM
```
purerosefallen, line 1487: How do you make sure the `code+event` thing does not lead into a conflict?

I suggest we use bitwise operations to to generate this event code.

purerosefallen: However, it's not that necessary to generate another code for the actual event code. That's what the script should maintain.

In case when we should register multiple events, we could do as what it looked like before, as follows:

```lua
local g=Group.CreateGroup()
aux.RegisterMergedDelayedEvent_ToSingelCard(c,59071624,EVENT_TO_GRAVE,g)
aux.RegisterMergedDelayedEvent_ToSingelCard(c,59071624,EVENT_REMOVE,g)

-- use EVENT_CUSTOM+59071624 in the following effects
```

To make it even simpler, we could make it like:

```lua
aux.RegisterMergedDelayedEvent_ToSingelCard(c,59071624,{EVENT_TO_GRAVE,EVENT_REMOVE},g)
```

c1688285.lua
kiritosoft: changed

utility.lua
kiritosoft: changed to English

utility.lua
kiritosoft: update to solve

c59071624.lua
```diff
@@ -24,11 +24,12 @@ function c59071624.initial_effect(c)
 	e2:SetOperation(c59071624.spop1)
 	c:RegisterEffect(e2)
 	--spsummon2
+	local custom_code=aux.RegisterMergedDelayedEvent_ToSingleCard(c,id,{EVENT_TO_GRAVE,EVENT_REMOVE})
```
Wind2009-Louse, line 27:
```suggestion
	local custom_code=aux.RegisterMergedDelayedEvent_ToSingleCard(c,59071624,{EVENT_TO_GRAVE,EVENT_REMOVE})
```

Wind2009-Louse (changes requested):

c66399653.lua
```diff
@@ -9,19 +9,17 @@ function c66399653.initial_effect(c)
 	e1:SetOperation(c66399653.activate)
 	c:RegisterEffect(e1)
 	--equip
+	local custom_code=aux.RegisterMergedDelayedEvent_ToSingleCard(c,id,{EVENT_SUMMON_SUCCESS,EVENT_SPSUMMON_SUCCESS})
```
Wind2009-Louse, line 12:
```suggestion
	local custom_code=aux.RegisterMergedDelayedEvent_ToSingleCard(c,66399653,{EVENT_SUMMON_SUCCESS,EVENT_SPSUMMON_SUCCESS})
```

Wind2009-Louse (changes requested):

salix5:
@Wind2009-Louse 
@kiritosoft 
cf0f1567d84e405ed4b9663312e864155ad2dabf
可能的話還是希望先確認Visual Studio Code的語法檢查有沒有問題

PR #2593, opened 2024-07-15
update "equipped by self effect" of Allure Queen

salix5:
For the effect of 
Allure Dance/c68316358
Allure Palace/c31322640


mercury233:
it is better to use card password

PR #2630, opened 2024-08-10
Update c61650133.lua

VaHReX:
Added a missing string, when activating the effect of banishing 1 card from the hand it showed the cost string instead, the string of banishing 1 card from the hand was added


mercury233:
It is using string 0 in SelectYesNo

VaHReX:
> It is using string 0 in SelectYesNo

Yes, I made the change because when the "Banish 1 card from hand" effect was chained, the string it showed you was "Pay 500LP to keep the card."

PR #2664, opened 2024-09-14
Fix 天盃龍 IsSynchroSummonable_check

Crescent0v0:
Use<code>Card.IsSynchroSummonable</code> with parameter <code>mg</code> in this way cause wrong check.
With change from core,it's no need to use  <code>c:IsStatus(STATUS_SUMMONING)</code>.
So    deleted.

by the way
天盃龍パイドラ & 天盃龍ファドラ keep using themselves as truner in <code>Card.IsSynchroSummonable</code> may cause bug in future.


salix5:
ded6d0232e132b88f0dffcc998cbbe3e628543f8
刪除函數時要將全部引用的地方都刪除

PR #2665, opened 2024-09-15
Add Mimighoul scripts

VaHReX:


salix5:
#2628 
For some reasons, recent card scripts are on different branches.

PR #2708, opened 2024-10-21
Revert "add param to initial_effect (#2702)"

mercury233:
This reverts commit 1ccf0bc57692c6d08800a4c7cfa46554f0e1ab44.


salix5:
Keeping Lua script correct is very difficult.
I think adding 1 row to check initial_effect is necessary.

mercury233:
I still don't think we should create difficulty to add those hint

purerosefallen:
~~We should change DataEditorX and the vscode plugin~~

salix5:
Checking the script by Lua language server has some disadvantages, but it is the best I can do.
If we (you?) cannot improve the script check, please do not make it even harder.

Lua language server檢查腳本有其缺點，但這已經是我所知道最好的方法了。
如果無法改進Lua腳本檢查，至少不要讓他變得更加困難。

mercury233:
I still don't think we should create difficulty to add those hint

salix5:
What kind of difficulties?
If script writers in the future does not want to add this line, they can omit this.

mercury233:
What if script writer don't want to add this line **now**?

salix5:
Then don't add this line **now**.
Adding this line or not totally depends on the script writers.
I did not**force** any script writers to add this line.

salix5:
@mercury233 
The future of this repo counts on you from now on.

PR #2745, opened 2024-11-16
Update c68473226.lua

VforViolin:
Fix Hardened Armed Dragon protection effect


salix5:
fixed in Fluorohydride/ygopro-core#661

PR #2793, opened 2024-12-26
fix RegisterMergedDelayedEvent_ToSingleCard

YiLong-jin:
This modification addresses two issues:

1.In the case of special summoning "クシャトリラ・アライズハート" using the effect of "RUM－ソウル・シェイブ・フォース," it triggered the effect of "幻影騎士団ラスティ・バルディッシュ" to destroy itself, after which the effect of "クシャトリラ・アライズハート" would not activate. 
2.The effects triggered by cards in hand when creating a point in time using this function cannot activate if they are triggered after the point in time when the cards are added to hand. At present, there is no hand card to induce effects using this function, if you want to test it needs to temporarily change such cards to call this function to cooperate with the test.

### DIFF
--- a/utility.lua
+++ b/utility.lua
@@ -1675,7 +1675,7 @@ end
 function Auxiliary.ThisCardMovedToPublicResetCheck_ToSingleCard(e,tp,eg,ep,ev,re,r,rp)
 	local c=e:GetOwner()
 	local g=e:GetLabelObject()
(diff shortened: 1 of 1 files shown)


Wind2009-Louse:
Current bug:
When you remove ``Despian Tragedy`` and set ``Branded Regained``, then activate the ``Branded Regained``, the ``Despian Tragedy`` can be recycled.

PR #2803, opened 2024-12-27
Fix Marincess Snow

Sonic714:
Fixed an issue where Marincess Snow's immunity effect is not reset when control switches.


Wind2009-Louse:
If the monster's controller switches again, should the immunity be recoverd?

Sonic714:
> If the monster's controller switches again, should the immunity be recoverd?

No, it would never regain immunity since its controller switched.

Wind2009-Louse:
```
なお、その後、そのモンスターのコントロールが再び自分に戻ったとしても、『この効果で特殊召喚したモンスターは自分フィールドに存在する限り、相手の効果を受けない』効果が再び適用される事はありません。
```

PR #2817, opened 2025-01-12
Update c63198739.lua

VforViolin:
Fix Standby Phase effect


VforViolin:
Nvm, I just realized its a translation error by Yugipedia to indicate it happens every Standby Phase, not just *your* standby.

PR #2841, opened 2025-02-09
Update c30888983.lua

VforViolin:
Fix to prevent activation when flip summoning a monster, while no other monsters are on the field with the same type.

Requires this commit
https://github.com/Fluorohydride/ygopro-scripts/pull/2840


mercury233:
Maybe we should update `card::is_treated_as_not_on_field()`

PR #2851, opened 2025-02-11
change to FusionProcMixRep for some fusion procs

purerosefallen:
FusionProcMixRep has better performance


purerosefallen:
Use #2852

mercury233:
Please stop committing untested codes

PR #2852, opened 2025-02-11
add AddFusionProcUltimate

purerosefallen:
The one-for-all fusion procedure, improving performance, saying goodbye to ygopro freeze for cards like Chimeratech Overdragon.

Cards not on field are treated topology-equalized, so they would not be checked recursively, increasing performance in cases of Chain Material, or having lots of cards in hand or grave.

### DIFF
--- a/c93053159.lua
+++ b/c93053159.lua
@@ -1,17 +1,8 @@
 --白の枢機竜
 local s,id,o=GetID()
 function s.initial_effect(c)
-	aux.AddCodeList(c,68468459)
-	aux.AddMaterialCodeList(c,68468459)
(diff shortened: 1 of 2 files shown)


purerosefallen:
DO NOT DELETE BRANCH

purerosefallen:
> #2409
> 
> > Temporary branch `test`
> > regenerated periodically
> 
> I have said that at the time I created branch `test` in 2024. The only purpose of that branch is deleting and re-creating at any time. What is wrong with that?

I hope this branch would not be deleted too soon, hoping someone finishing this idea.

PR #2853, opened 2025-02-13
update 盆回し

purerosefallen:
it should reset when no longer valid, to improve performance


mercury233:
The program may crash in my test

PR #2863, opened 2025-02-19
Update c64664373.lua

VforViolin:
Fix to prevent use on Normal monsters, or monsters who are already negated.

https://db.ygoresources.com/card#19910

"It cannot be activated by targeting normal monsters or monsters whose effects are nullified."

### DIFF
--- a/c64664373.lua
+++ b/c64664373.lua
@@ -14,7 +14,7 @@ function s.initial_effect(c)
 	c:RegisterEffect(e1)
 end
 function s.tgfilter(c,e,tp)
-	return c:IsFaceup() and not c:IsType(TYPE_TOKEN) and not c:IsDisabled() and c:IsType(TYPE_EFFECT) 
+	return aux.NegateEffectMonsterFilter(c) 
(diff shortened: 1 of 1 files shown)


c64664373.lua
```diff
@@ -14,7 +14,7 @@ function s.initial_effect(c)
 	c:RegisterEffect(e1)
 end
 function s.tgfilter(c,e,tp)
-	return c:IsFaceup() and not c:IsType(TYPE_TOKEN)
+	return c:IsFaceup() and not c:IsType(TYPE_TOKEN) and not c:IsDisabled() and c:IsType(TYPE_EFFECT)
```
Wind2009-Louse, line 17:
```suggestion
	return aux.NegateEffectMonsterFilter(c) 
```

Wind2009-Louse (changes requested):

c64664373.lua
VforViolin: While tokens cannot be given effects currently, would it not be worth keeping the token check as well incase this changes in the future?

Wind2009-Louse: > While tokens cannot be given effects currently, would it not be worth keeping the token check as well incase this changes in the future?

Tokens can be given effect by cards like **Ryu-Ge Realm - Dino Domains**.

PR #2882, opened 2025-03-15
Fix fusion effect

Wind2009-Louse:
* Replace aux.FCheckAdditional to FGoalCheckAdditional.
* Add Card.IsImmuneToEffect check in target.
* Fix grammar mistakes about Chain Material.


purerosefallen:
> Replace aux.FCheckAdditional to FGoalCheckAdditional.

PR #2887, opened 2025-03-22
Fix set card shuffling

VforViolin:
Fix the shuffling of set cards, set by Turbulence, Toy Box and Divine Serpent Apophis, as per rulings. The opponent is not supposed to be shown what card is set where.

https://db.ygoresources.com/card#6868

https://db.ygoresources.com/card#17997


mercury233:
It should be hidden to the opponent after https://github.com/Fluorohydride/ygopro-core/pull/165

VforViolin:
In-game on both DNX and Omega the opponent sees where the cards are set, and they dont seem to ever change location without the shuffle, which is why I had added it here. Perhaps that core change isnt working?

mercury233:
I just did some test with _Multirole_, it is working in YGOPro, the opponent will see the cards set from grave in random order, following by a shuffle effect, then the card names will be invisible and the order will be different.

VforViolin:
I guess I'll have to do more testing to see why theres a discrepancy on our end then.

PR #2914, opened 2025-04-23
fix ambiguous function name `tdcfop` to `ContactFusionSendToDeck`

drunkenQCat:
it's so unique that `tdcfop` is the only function out of naming convention. And also the function is usually used in Gladiator fusion monsters. After reviewing the Gladiators' code, I decided to rename them to `ContactFusionSendToDeck`, as an option of contact fusion.


mercury233 (approved):

purerosefallen:
We may add a compat name, and deprecate it in future versions.

PR #2921, opened 2025-05-05
Update Fortune Lady Earth

zengsxing:
修复命运女郎·厄希(82971335)在部分情况下无法检测到等级变化的问题


purerosefallen:
we may have to fix `EVENT_LEVEL_UP` in core

PR #2928, opened 2025-05-14
Fix Mitsurugi summon effects

VforViolin:
Fix summon effects so Ghost Belle can respond to them.


purerosefallen:
Better to add another `SetOperationInfo`

mercury233:
> SetOperationInfo

means determined action

PR #2957, opened 2025-06-08
fix: "Proton Blast" and "Lucky Chance"

ykxakks:
Fix the condition of the effect of resetting results of coin tosses in "Proton Blast".
For coin_toss effects, we should define the last parameter in Duel.SetOperationInfo(0,CATEGORY_COIN,nil,0,PLAYER_ALL,1) to be the number of coin tosses for each player when the player parameter is set to PLAYER_ALL. In this case, we should calculate the ct in condition of "Proton Blast" by ct*2 if the cp is PLAYER_ALL.

There is currently one card that uses Duel.SetOperationInfo for CATEGORY_COIN and cp is PLAYER_ALL: "BIG Win!?"(84677654), so the last parameter is changed to 1 according to this new definition. It is checked that "Proton Blast" can still be activated when "BIG Win!?" is activated.

For "Lucky Chance": it uses the same logic as "Proton Blast" to calculate the total number of coin tosses. Currently it works fine since no monster cards do Duel.SetOperationInfo for PLAYER_ALL, but it should also be fixed. Since "Lucky Chance" can only be activated when the total count is exactly 1, `cp~=PLAYER_ALL` should be a condition that must be satisfied.


ykxakks:
Note: the bug that "Proton Blast" cannot make both coin tosses of "BIG Win!?" to be heads is not fixed in this PR.

ykxakks:
the bug that ""Proton Blast" cannot make both coin tosses of "BIG Win!?" to be heads" is fixed in [4d739b9](https://github.com/Fluorohydride/ygopro-scripts/pull/2957/commits/4d739b96d5729f8f550d1cebad4fad96562811e6), where we remove the SetCountLimit(1) in coinop1 so it can be applied to all coin tosses in the target effect.

PR #2977, opened 2025-07-01
Fix Timaeus the United Magidragon

VforViolin:
Fix fusion material procedure so substitutes like King of the Swamp can be used.


amcsi:
Note that at least for Omega this is fixed already in theory, as per this comment in Duelists Unit discord:

https://discord.com/channels/253945062625378304/285844671836717057/1389611485013213184

VforViolin:
Yes, because I applied this change locally to Omega.

amcsi:
Oh sorry, I thought this was an issue, not a PR

PR #2983, opened 2025-07-16
Update Kashtira Fenrir

VforViolin:
Add an additional check to prevent Fenrir from triggering when Argostars Lightning Tydeu activates the summon effect in the same chain as Tydeu (the card) activates.


mercury233:
I can't reproduce it with the current YGOPro.

PR #3000, opened 2025-08-11
Fix facedown special summon effect

Wind2009-Louse:
When applied ``糾罪巧``'s splimit, some effect doesn't run normally.


Wind2009-Louse:
Fixed by updating ``糾罪巧``.

PR #3023, opened 2025-09-12
Fix: Prevent targeted Spell/Trap from being destroyed etc if its activation is negated (Unchained)

vuryleo:
This fixes an incorrect behavior where a targeted Spell/Trap card would still be destroyed even if its activation was negated.

Details:

According to official rulings, when the activation of a Spell/Trap card is activation negated, the card is no longer considered to be on the field. Effects that were specifically targeting that card are no longer applied.

This PR corrects the scripts so that Spell/Trap cards with negated activations are not erroneously destroyed etc.

The method is to add another `IsOnField` check while resolving. Cards get activation negated will not be consider on the field anymore.

The cards are too many, i will split them across PRs.

For this PR we have:

* `サイクロン`
* `破械童子アルハ`
* `破械童子ラキア`
* `破械雙王神ライゴウ`
* `破械神の慟哭`
* `破械唱導`
* `破械神シャバラ`
* `破械神シュヤーマ`

Do you think it's the right approach? I won't open more PR before confirming it should work like this.
@purerosefallen perhaps. The full list will contains about ~800 cards.

Actually I would suggest that we release them in 888 gracefully, and if nothing broken we can merge it here.


Wind2009-Louse:
I prefer to change the core to fix this, rather than update tons of cards without knowing if any are still missing.

But this can be a workaround.

vuryleo:
which method from core do you think could cover all scenario?

PR #3041, opened 2025-10-01
Update Synchro Fellows and D/D/D Sky King Zeus Ragnarok

VforViolin:
Fix both scripts to no longer produce a "[string "utility"]1161: attempt to call a nil value (local 'f2')" error on Omega and DNX. I would imagine these may also produce similar errors on ygopro.

### DIFF
--- a/c30998403.lua
+++ b/c30998403.lua
@@ -74,11 +74,14 @@ end
 function s.cfilter2(c)
 	return c:IsSetCard(0xae) and c:IsAbleToRemoveAsCost()
 end
+function s.fselect(g,c,tp)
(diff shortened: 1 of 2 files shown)


mercury233:
can't reproduce, but I guess it is something wrong with the lua engine

mercury233:
> can't reproduce, but I guess it is something wrong with the lua engine

https://github.com/Fluorohydride/ygopro-scripts/pull/3202

It do be a [limitation](http://lua-users.org/wiki/StoringNilsInTables) of the the lua engine.

VforViolin:
Honestly, Im just glad to know that theres a potential solution other than what I had suggested above, which *works*, but doesnt address the actual underlying issue.

PR #3084, opened 2025-12-12
王宮の弾圧: Fix activation while Set

vuryleo:
# Description

This PR fixes an issue where 王宮の弾圧 could not be activated while Set to negate a Special Summon.

According to correct rulings, 王宮の弾圧 can be activated directly from the Set state in response to a Special Summon. The previous implementation incorrectly required the card to be face-up, preventing activation at the proper timing.

# Changes

* Allow 王宮の弾圧 to activate while Set
* Fix Special Summon negation timing to match correct behavior

For review, please use the GitHub web diff between these two commits:
https://github.com/Fluorohydride/ygopro-scripts/compare/527bb073bba2ca383ddd00e8a993d13157547f99...5754caaa01e68ec51669b2840d3205d5d09e33e9

### DIFF
--- a/c93016201.lua
+++ b/c93016201.lua
@@ -1,77 +1,79 @@
--- 王宮の弾圧
-local s, id, o = GetID()
+--王宮の弾圧
+local s,id,o=GetID()
 function s.initial_effect(c)
-    -- Activate
-    local e1 = Effect.CreateEffect(c)
-    e1:SetType(EFFECT_TYPE_ACTIVATE)
-    e1:SetCode(EVENT_FREE_CHAIN)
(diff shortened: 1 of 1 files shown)


vuryleo:
image

mercury233:
https://github.com/Fluorohydride/ygopro-core/blob/2d2e322d68c5ea7b858669605cb6dd672b0a939b/operations.cpp#L2906

`EVENT_SPSUMMON` set `skip_freechain` to 1 when checking the event, it is designed that `EVENT_FREE_CHAIN` effects can't be actived in chain 1 during the summoning step. The above changement of continuous trap didn't consider this so it break that card.

PR #3120, opened 2026-02-02
原石の鳴獰: Lock declared names after successful resolution until opponent’s End Phase

vuryleo:
# Description
This PR fixes an issue where 原石の鳴獰 could repeatedly resolve using the same declared monster name within the same turn window.

# Intended behavior:
* The card can be activated and a monster name can be declared as normal.
* Once a declared code has successfully resolved, that code is treated as used.
* While a code is used:
    - it cannot be declared again on subsequent activations, and
    - if referenced again at resolution time (e.g. within the same chain), the effect does not resolve.
* The restriction persists until the end of the opponent’s turn, at which point the used-code state is cleared. 

# Testing
## Scenario 1
* Activate 原石の鳴獰, declare a valid Normal Monster code, and let it resolve.
* Activate again and confirm the same code cannot be declared.
* After the opponent’s End Phase, confirm the code can be declared again.

## Scenario 2
* With an empty field, activate 原石の鳴獰 and declare a valid Normal Monster code.
* Chain another 原石の鳴獰 declaring the same code; resolve without Special Summon.
* Resolve the first 原石の鳴獰 and confirm no Special Summon occurs.

### DIFF
--- a/c92501449.lua
+++ b/c92501449.lua
@@ -22,7 +22,27 @@ function s.initial_effect(c)
 	e2:SetTarget(s.rmtg)
 	e2:SetOperation(s.rmop)
 	c:RegisterEffect(e2)
+	if not s.global_check then
+		s.global_check=true
+		-- reset announced codes at opponent's End Phase
+		local ge=Effect.CreateEffect(c)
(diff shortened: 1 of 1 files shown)


fallenstardust:
merge？

PR #3126, opened 2026-02-07
Fix destroy replacement filters: enforce IsImmuneToEffect and snapshot destroy group

vuryleo:
This PR updates the destroy-replacement implementation for 六花聖カンザシ, エヴォルド・フォリス and 魔界特派員デスキャスター etc to correctly handle destruction group snapshotting and release validation during EFFECT_DESTROY_REPLACE.

This PR updates ~30 card scripts that use EFFECT_DESTROY_REPLACE to ensure the replacement target is always legal, with particular focus on enforcing effect immunity checks (not c:IsImmuneToEffect(e)) when selecting the replace candidate.

# Testing notes
Confirmed replacement cannot select:
* cards to be destroyed
* cards immune to the replacing effect

### DIFF
--- a/c17228908.lua
+++ b/c17228908.lua
@@ -93,7 +93,7 @@ function c17228908.repfilter(c,tp)
 end
 function c17228908.desfilter(c,e)
 	return c:IsRace(RACE_DINOSAUR) and c:IsDestructable(e)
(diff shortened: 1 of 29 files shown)


vuryleo:
i only manually tested 六花聖カンザシ, エヴォルド・フォリス and 魔界特派員デスキャスター

PR #3185, opened 2026-05-14
Fix フィリアス・ディアベル

Tad-Done:
Fix the issue where  ⌈フィリアス・ディアベル⌋ would prevent ⌈ディアベル⌋ monsters from being affected by other continuous attack-changing effects under specific situations.


Wind2009-Louse:
Can you provide a puzzle/replay that can reproduce this scenario?

Tad-Done:
[05-13「19：26：05」.zip](https://github.com/user-attachments/files/28228690/05-13.19.26.05.zip)
Turn 67, when Diabellstar Vengeance was flipped up, its atk should have been 3400 because of D.D. Post. Cannot reproduce the scenario now. I suppose this pull request can be closed.

PR #3211, opened 2026-06-16
fix Mulcharmy monsters

mercury233:
According to the [ruling](https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=7813&keyword=&tag=-1&request_locale=ja), the activation count limit of _Mulcharmy_ monsters' effects is still counted even if the activation is negated. But `ACTIVITY_CHAIN` reduces its count when the activation is negated.

### DIFF
--- a/c42141493.lua
+++ b/c42141493.lua
@@ -1,6 +1,7 @@
 --マルチャミー・フワロス
 local s,id,o=GetID()
 function s.initial_effect(c)
+	aux.EnableMulcharmyGlobalCheck()
 	--draw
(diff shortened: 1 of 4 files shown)


mercury233:
This change removes the current restriction that apply to **other** Mulcharmy cards after enough Mulcharmy cards have already been activated. Instead, it relies solely on each Mulcharmy card's own activation restriction, assuming that all future Mulcharmy cards will share the same activation conditions.

PR #3215, opened 2026-06-21
fix abuse of aux.AND & NecroValleyFilter

mercury233:
- `aux.AND`: Will pass varargs to ALL functions, it means, `aux.AND(Card.IsRelateToChain,Card.IsType)...TYPE_MONSTER` will pass `TYPE_MONSTER` to `Card.IsRelateToChain`.
- ~~`aux.NecroValleyFilter`: Not necessary for effects that target cards in grave and set operation info correctly. _Necrovalley_ will negate those effects in `c47355498.disop`.~~ It is wrong, see https://github.com/Fluorohydride/ygopro-scripts/pull/3224

- _Dracotail Shaulas_: Fix to use same ruling as _Ecclesia and the Dark Dragon_.

### DIFF
--- a/c12210097.lua
+++ b/c12210097.lua
@@ -46,7 +46,7 @@ function s.tdtg(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
 	Duel.SetOperationInfo(0,CATEGORY_TODECK,g,g:GetCount(),0,0)
 end
 function s.tdop(e,tp,eg,ep,ev,re,r,rp)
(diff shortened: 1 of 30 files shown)


mercury233:
There are way more abuse of `aux.NecroValleyFilter()(tc)`. Will clean them up later.

mercury233:
我很抱歉，之前我做出的关于 `aux.NecroValleyFilter` 的修改是错误的。

I must apologize for the wrong modification about `aux.NecroValleyFilter` I did some day ago.

https://github.com/Fluorohydride/ygopro-scripts/pull/3224

PR #3217, opened 2026-06-22
cleanup timing hints

mercury233:


mercury233:
Although this conflicts with the intent of https://github.com/Fluorohydride/ygopro-scripts/commit/b1e1b6ad7d608814e4fb83cd9335455adf22fa45, I'm temporarily keeping some `TIMING_DRAW_PHASE` flags. Certain clients rely on these hints to activate these cards during the first-turn Draw Phase (to avoid triggering _Droll & Lock Bird_). These `TIMING_DRAW_PHASE` flags will be removed in the future.

PR #3223, opened 2026-06-26
fix typos HINGMSG_NUMBER, HINGMSG_LVRANK

purerosefallen:
done by @salix5 

I force-pushed it back because it's an expansion-breaking change affecting prod

and compat vars also should be added


salix5:
Please don't force push under any circumstances,  even if the commit is wrong. 

If you believe it is wrong,  please revert that commit.

PR #3224, opened 2026-06-26
revert changes about NecroValleyFilter

mercury233:
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=14915&keyword=&tag=-1&request_locale=ja

我很抱歉，之前我做出的关于 `aux.NecroValleyFilter` 的修改是错误的。我产生了错误的记忆，以为关于 _王家长眠之谷_ 和 _幻变骚灵协议_ 的裁定不再适用，但实际上并没有发生。
_王家长眠之谷_ 适用中，即使 _幻变骚灵协议_ 也适用，幻变骚灵卡的效果不会被无效化，王谷仍然会使涉及墓地的部分不适用。所以需要在脚本中做额外的判断。

I must apologize for the wrong modification about `aux.NecroValleyFilter` I did some day ago. I mistakenly remembered that the ruling regarding _Necrovalley_ and _Altergeist Protocol_ had been rescinded, but that never actually happened.
When _Necrovalley_ is in effect, even if _Altergeist Protocol_ is also in effect and the effects of Altergeist cards won't be negated, _Necrovalley_ will still invalidates the Graveyard-related parts of those effects. So extra check must be done in the scripts.

### DIFF
--- a/c12210097.lua
+++ b/c12210097.lua
@@ -46,7 +46,7 @@ function s.tdtg(e,tp,eg,ep,ev,re,r,rp,chk,chkc)
 	Duel.SetOperationInfo(0,CATEGORY_TODECK,g,g:GetCount(),0,0)
 end
 function s.tdop(e,tp,eg,ep,ev,re,r,rp)
-	local g=Duel.GetTargetsRelateToChain()
+	local g=Duel.GetTargetsRelateToChain():Filter(aux.NecroValleyFilter(),nil)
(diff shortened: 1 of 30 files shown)


Wind2009-Louse:
The current ``aux.NecroValleyFilter`` cannot distinguish between the following two scenarios:
* The effect of a card is not negated by Necrovalley.
* The effect of a card is not affected by Necrovalley.

For the former, even if the effect is not negated, it still cannot affect cards in the Graveyard, whereas the latter can. The current change is merely a simple rollback and does not fully fix this issue.

PR #3239, opened 2026-07-19
Fix White Sunfish

VforViolin:
Fix summon effect to prevent targeting a monster in the GY if the matching monster on the field is facedown.


mercury233:
Creating a new function in the filter may significantly increase the time complexity.

PR #3241, opened 2026-07-19
Fix Genia of the Ring

VforViolin:
Fix special summon effect to allow you to target a monster and perform the summon even if its already a Spellcaster.


VforViolin:
https://db.ygoresources.com/card#22927

PR #3243, opened 2026-07-21
Destiny Board Bug Fix

sakusankun:
These bugs are fixed based on the OCG official ruling or based on interpretation of card text.
1. When a face-down "Spirit Message" card or "Destiny Board"  leaves the field of the controler of "Destiny Board" (Same for Destiny Board itself), but sent to deck, or hand (not public), the second effect of Destiny Board should not be triggered (based on official ruling).
2. When a "Spirit Message" card or "Destiny Board" leaves the opponent's field of the controler of "Destiny Board" (Which means the controler was changed), but sent to the controler's GY, the second effect of Destiny Board should not be triggered (based on card text: When any "Spirit Message" card or "Destiny Board" you control leaves the field/自分フィールドの「ウィジャ盤」または「死のメッセージ」カードがフィールドから離れた時に).


sakusankun:
QQ_1784699385840

PR #3247, opened 2026-07-30
HINGMSG_NUMBER, HINGMSG_LVRANK for compat

purerosefallen:
https://github.com/Fluorohydride/ygopro-scripts/pull/3225

Keep for compat for at least 1 release


mercury233 (approved):
I suggest removing these compatibility definitions, as well as `DEFENCE`, the next time we bump the version number. Until then, they can be added back.

PR #3275, opened 2026-09-06
Fix Priestess with Eyes of Blue

VforViolin:
Fix effect to only allow summoning if the target is shuffled into the deck. Previously, if the monster would be banished by "EFFECT_LEAVE_FIELD_REDIRECT", Priestess would still summon itself.

### DIFF
--- a/c36734924.lua
+++ b/c36734924.lua
@@ -68,7 +68,7 @@ end
 function c36734924.spop(e,tp,eg,ep,ev,re,r,rp)
 	local c=e:GetHandler()
 	local tc=Duel.GetFirstTarget()
-	if tc:IsRelateToEffect(e) and Duel.SendtoDeck(tc,nil,SEQ_DECKSHUFFLE,REASON_EFFECT)~=0 and c:IsRelateToEffect(e) then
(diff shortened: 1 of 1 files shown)


c36734924.lua
```diff
@@ -68,7 +68,7 @@ end
 function c36734924.spop(e,tp,eg,ep,ev,re,r,rp)
 	local c=e:GetHandler()
 	local tc=Duel.GetFirstTarget()
-	if tc:IsRelateToEffect(e) and Duel.SendtoDeck(tc,nil,SEQ_DECKSHUFFLE,REASON_EFFECT)~=0 and c:IsRelateToEffect(e) then
+	if tc:IsRelateToEffect(e) and Duel.SendtoDeck(tc,nil,SEQ_DECKSHUFFLE,REASON_EFFECT)~=0 and c:IsRelateToEffect(e) and tc:IsLocation(LOCATION_DECK) then
```
Wind2009-Louse: > 対象のモンスターをデッキ、**またはエクストラデッキ**に戻すことに成功した場合、『このカードを墓地から特殊召喚する』処理が行われます。

I think it should be ``tc:IsLocation(LOCATION_DECK+LOCATION_EXTRA)``

Wind2009-Louse (changes requested):

c36734924.lua
VforViolin: I agree.

